PR #6, opened 2021-03-31
Updated Github pages documentation publish URL references

VRamakrishna:


knarayan (approved):
Changes look good to me.

sanvenDev (approved):

VRamakrishna:
> 
> 
> Admittedly this is orthogonal to what this PR is about but I just happened to have a look at these files and the use of "here" everywhere jumped at me. You should never use "here" as the anchor for a link. Instead use some meaningful text.
> I can submit a PR to fix that.

@lehors Sure, please do that if you can. (Or make suggestions for me to fix.)

PR #42, opened 2021-04-21
Upgrade Node.js Dependency to Latest Version

sanvenDev:
1. Modified `package.json` to support Latest Node Js Version.
2. Removed `package-lock.json` from every node app folder as it used to create conflict when it was created by older version of node.
3. Tested using Node v15.14.1, fabric-fabric, fabric-corda, corda-fabric interop working on my VM.


sanvenDev:
Updated Getting Started Page as well.

VRamakrishna:
@sanvenDev As discussed, please test on different versions of node from 10 up to the latest so we can specify the right lower bound for the version in the documentation.

docs/docs/external/getting-started.md
```diff
@@ -118,7 +118,7 @@ The `fabric-cli` source code is located in the `tests/network-setups/fabric/dev/
 
 If you are using a Linux system, make sure that lib64 is installed.
 
-_Note_: The setup and running instructions below were tested with Nodei.js v10.16.3.
+_Note_: The setup and running instructions below were tested with Node.js v10.16.3 and v15.14.1.
```
VRamakrishna, line 121: @sanvenDev Can we say here that the setup and instructions were tested using various versions between v10.16.3 and v15.14.1?

Also, in the "Prerequisites" section above, can we say that any Node version between 10 and 15 (both inclusive) can be installed?

lehors: Node.js v10 is ending this Friday, we should not encourage people to use it. v12 is the minimum recommended.

VRamakrishna: Right, in fact this issue was created precisely because support for v10 is ending. @sanvenDev Please change the instructions to state that we support v11 to v15.

VRamakrishna (approved):

PR #44, opened 2021-04-30
Asset exchange base class for Fabric chaincodes

VRamakrishna:
Implemented a set of library functions that can be exercised by any Fabric application chaincode that is managing assets. Via the function API exported by this case class, the application chaincode can trigger asset locks and claims (processed eventually by the Fabric Interop Chaincode) for a cross-network atomic swap. Presently, only the HTLC mechanism is supported for such swaps. 

Also defined common structures (protobufs) relevant to the asset exchange protocol.


core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks_test.go
```diff
@@ -0,0 +1,1510 @@
...
+    }
+    if function == "UnlockFungibleAsset" {
+        expectedKey := args[0] + ":" + args[1]
+        expectedVal := string(caller) + ":" + args[2]
+        if cc.fungibleAssetLockMap[expectedKey] == "" {
+            return shim.Error(fmt.Sprintf("No asset of type %s and Units %s is locked", args[0], args[1]))
+        } else if cc.fungibleAssetLockMap[expectedKey] != expectedVal {
+            return shim.Error(fmt.Sprintf("Cannot unlock asset of type %s and Units %s as it is locked by %s for %s", args[0], args[1], string(caller), args[2]))
+        } else {
+            delete(cc.fungibleAssetLockMap, expectedKey)
```
knarayan, line 136: We have to reset the `fungibleAssetUnlockedCount` ideally here (e.g., `cc.fungibleAssetUnlockedCount[args[0]] = cc.fungibleAssetUnlockedCount[args[1]] + int(args[2])`).

knarayan (changes requested):
@VRamakrishna, request you to go over the inline comments.

knarayan (approved):

common/interop-protos/common/asset_locks.proto
```diff
@@ -0,0 +1,46 @@
...
+    DURATION = 1;
+  }
+  TimeSpec timeSpec = 3;
+}
+
+message AssetClaimHTLC {
+  bytes hashPreimage = 1;
+}
+
+message AssetAgreement {
```
ermyas, line 34: Perhaps calling this `AssetExchangeAgreement` might be more descriptive

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+        log.Error(err.Error())
+        return false, err
+    }
+    iccResp := stub.InvokeChaincode(am.interopChaincodeId, [][]byte{[]byte("LockAsset"), assetAgreementBytes, lockInfoBytes}, "")
+    fmt.Printf("Response from Interop CC: %+v\n", iccResp)
+    if iccResp.GetStatus() != shim.OK {
+        return false, fmt.Errorf(string(iccResp.GetPayload()))
+    }
+    lockExpiryTimeSec := int64(lockInfoHTLC.ExpiryTimeMillis)/1000
+    lockExpiryTimeNano := (int64(lockInfoHTLC.ExpiryTimeMillis) - (lockExpiryTimeSec * 1000)) * (1000 * 1000)     // Convert milliseconds to nanoseconds
```
ermyas, line 135: Do we really need this (pseudo) nanosecond resolution time here? Given we are not operating at that time resolution should we just pass 0 to the nanosecond arg of `time.Unix()` instead?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+    assetAgreementBytes, err := proto.Marshal(assetAgreement)
+    if err != nil {
+        log.Error(err.Error())
+        return false, err
+    }
+    lockInfoBytes, err := proto.Marshal(lockInfoHTLC)
+    if err != nil {
+        log.Error(err.Error())
+        return false, err
+    }
```
ermyas, line 197: There is a significant amount of redundant boilerplate in how we perform validations of args in each function, which reduces readability and probably maintainability. I suggest we refactor to consolidate common validation logic in separate reusable helper functions and minimize validation boilerplate in the core API methods.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+        errorMsg = "Empty asset ID"
+        log.Error(errorMsg)
+        return false, errors.New(errorMsg)
+    }
+    if len(assetAgreement.Recipient) == 0 {
+        errorMsg = "Empty lock recipient"
+        log.Error(errorMsg)
+        return false, errors.New(errorMsg)
+    }
+    if len(lockInfoHTLC.Hash) == 0 {
```
ermyas, line 114: perhaps we can bring this next to other validation checks on `lockInfoHTLC` (i.e. move to line 98)

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+        infoMsg = "Empty locker; assuming caller"
+        log.Info(infoMsg)
+        assetAgreement.Locker = myself
+    }
+    if assetAgreement.Recipient == assetAgreement.Locker {
+        errorMsg = "Invalid query: locker identical to recipient"
+        log.Error(errorMsg)
+        return false, errors.New(errorMsg)
+    }
+    if assetAgreement.Recipient != myself && assetAgreement.Locker != myself {
```
ermyas, line 250: Is this something we want to enforce at this level, or this is an application contract level decision? Could we envision legitimate scenarios where third parties (auditors etc.) are entitled to verify this information?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+    if len(assetAgreement.Locker) == 0 {
+        infoMsg = "Empty locker; assuming caller"
+        log.Info(infoMsg)
+        assetAgreement.Locker = myself
+    }
+    if assetAgreement.Recipient == assetAgreement.Locker {
+        errorMsg = "Invalid query: locker identical to recipient"
+        log.Error(errorMsg)
+        return false, errors.New(errorMsg)
+    }
```
ermyas, line 249: Should a user not be able to query whether an asset of a specific ID is locked for/or by them, without specifying the counterparty?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+}
+
+
+// Ledger transaction (invocation) functions
+
+func (am *AssetManagement) AddFungibleAssetCount(stub shim.ChaincodeStubInterface, assetType string, numUnits int) (bool, error) {
+    var errorMsg string
+
+    if len(am.interopChaincodeId) == 0 {
+        errorMsg = "Interoperation chaincode ID not set. Run the 'Configure(...) function first."
```
ermyas, line 47: looks like the closing single quote is missing after `Configure()` ... comment applies to all other places where this message is duplicated in other functions

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+    }
+
+    if (lockInfo.LockMechanism != common.LockMechanism_HTLC) {
+        log.Errorf("Unsupported lock mechanism: %+v", lockInfo.LockMechanism)
+        return false, fmt.Errorf("Unsupported lock mechanism: %+v", lockInfo.LockMechanism)
+    }
+    lockInfoHTLC := &common.AssetLockHTLC{}
+    if len(lockInfo.LockInfo) == 0 {
+        log.Error("Empty lock info")
+        return false, fmt.Errorf("Empty lock info")
```
ermyas, line 87: for consistency should we introduce a formatted string variable similar to `errorMsg` used in all other instances (e.g. line 75), for the messages that are used across multiple statements? comment applies to the various places a similar pattern is used.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+
+// AssetManagement provides functions to manage asset locks by calling suitable functions in the interop contract
+type AssetManagement struct {
+    shim.Chaincode
+    interopChaincodeId string
+}
+
+
+// Utility functions
+func (am *AssetManagement) Configure(interopChaincodeId string) {
```
ermyas, line 36: Do we assume that access controls around who would be allowed to call this function are controlled by the application? If not, I assume this introduces a point of vulnerability if not controlled? Is there a strong reason this should be dynamic as opposed to set once during initialization?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+    am.interopChaincodeId = interopChaincodeId
+}
+
+
+// Ledger transaction (invocation) functions
+
+func (am *AssetManagement) AddFungibleAssetCount(stub shim.ChaincodeStubInterface, assetType string, numUnits int) (bool, error) {
```
ermyas, line 43: it is less likely this will be a notable issue but it might be safer to replace the use of `int`, whose size (32 or 64) is platform-dependent (which could cause non-determinism), with something more fixed .. ideally `int64`, to avoid overflows

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -0,0 +1,828 @@
...
+    var errorMsg string
+
+    if len(am.interopChaincodeId) == 0 {
+        errorMsg = "Interoperation chaincode ID not set. Run the 'Configure(...) function first."
+        log.Error(errorMsg)
+        return false, errors.New(errorMsg)
+    }
+
+    if len(assetType) == 0 {
+        errorMsg = "Empty asset type"
```
ermyas, line 53: I guess the Go convention around error messages is that they are not capitalized and do not end in punctuations, because it is not known how the caller will embed that message in its logs (e.g. in the middle of a sentence). Might not hurt to follow that convention here as well. Comment applies to all other error message strings ...

ermyas (changes requested):
mostly minor comments

common/interop-protos/common/asset_locks.proto
VRamakrishna: Yes, I'll change it.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: We lose some precision if we just retain the seconds part, assuming the time passed in the argument is indeed at millisecond granularity. The nanosecond field contains whatever balance remains after extracting the seconds value.

If we believe that we can't get more granular than seconds, we should just change the field to `ExpiryTimeSecs` rather than `ExpiryTimeMillis`.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Whether or not argument validation separation from the main function makes the code more readable is up for debate in my opinion.

I do agree that boilerplate code redundancy is not a great thing, but different functions perform validations on different parameter combinations, so I don't believe we will end up creating a small set of helper functions. Rather, each helper function will likely check for "emptiness" of a single parameter.

But I'll review the code once again to determine if such refactoring is useful.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: I think you are right. This is too restrictive. I'll remove the check.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Right. I'll fix it.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Yes, they should. I'd envisioned the use of a `*` for this purpose, but that will be interpreted by the Fabric Interop CC. @knarayan please take note.

In fact, following on the previous comment, a user should be able to check whether an asset is locked period, regardless of whether it is a party to the agreement. I'll review the code to determine if all bases are covered.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: This code is not going to run in a separate module. It's part of any application chaincode that wishes to inherit and use it. So there is no way to initialize its state (i.e., the interop chaincode ID) other than through the application chaincode (typically in the init function).

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: I think `int` is pretty much guaranteed to be 32 bits, isn't it? Most languages other than Go use `long` for 64-bit integers. I can change this to `int32` to make it explicit. But why do you think it should be `int64`? 32 bits already allow us to express numbers in excess of 2 trillion.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: OK, I didn't realize this. I'll fix it.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks_test.go
```diff
@@ -0,0 +1,1510 @@
...
+    }
+    if function == "LockFungibleAsset" {    // We are only going to lock once or twice in each unit test function, so bookkeeping doesn't need to be thorough
+        assetAgreement := &common.FungibleAssetAgreement{}
+        _ = proto.Unmarshal([]byte(args[0]), assetAgreement)
+        key := assetAgreement.Type + ":" + strconv.Itoa(int(assetAgreement.NumUnits))
+        val := string(caller) + ":" + assetAgreement.Recipient
+        if cc.fungibleAssetUnlockedCount[assetAgreement.Type] < int(assetAgreement.NumUnits) {
+            return shim.Error(fmt.Sprintf("Requested %d units of asset type %s; only %d available", assetAgreement.NumUnits, assetAgreement.Type, cc.fungibleAssetUnlockedCount[assetAgreement.Type]))
+        }
+        cc.fungibleAssetLockMap[key] = val
```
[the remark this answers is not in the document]
VRamakrishna: Right, we'll need to cover all corner cases comprehensively in the real Fabric Interop CC, but this one is meant to be a dummy just for unit testing purposes, so the logic here is sufficient.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks_test.go
VRamakrishna: Same answer as above.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
ermyas: - The size of `int` in Go is platform-dependent, and not guaranteed to be 32 bit: https://golangbyexample.com/go-size-range-int-uint/
- I think 32 bits would be up to 2 billion? not trillion

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
ermyas: I don't think we will be able to enforce these contracts at millisecond second resolution either, so I am not sure whether any meaningful amount of precision is lost here for our purposeses. I agree with changing the field to `ExpiryTimeSecs`

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks_test.go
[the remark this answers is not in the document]
lehors: I suggest a comment that explains just that so that other people don't wonder.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
ermyas: also, should we use unsigned int type instead?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks_test.go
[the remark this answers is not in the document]
VRamakrishna: Done

common/interop-protos/common/asset_locks.proto
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Right, it's over 2 billion. Still sufficiently large for a fungible asset count, no?

I think using `int` and flagging an error when a negative value is passed in OK. Why do you prefer `uint`?

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done. Here and in other places.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done in the base class. Fabric Interop CC needs to handle *.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
VRamakrishna: Done. Using `errors.New(....)` everywhere now.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
ermyas: The reason I think 2 Billion would be small for fungible tokens is that often what we are tracking with this number is the smallest possible denomination of a fungible asset (since this is an int and not a floating-point number). For instance, for Bitcoin that would be _Satoshi_, and for Ethereum that would be _Wei_. 2 Billion would not be enough to represent 22BTC (2,200,000,000 Satoshi) or even 1 ETH (1,000,000,000,000,000,000 wei)

ermyas: regarding `int` vs `uint`: my thinking here is why dynamically enforce through code what we can statically enforce through the type system? ... but I don't really mind either way in this context

VRamakrishna: Done. Changed everything to `uint64`.

PR #46, opened 2021-05-08
Interop-cc chaincode for asset management functions (Lock, Unlock, Claim, IsAssetLocked)

knarayan:


core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,310 @@
...
+		return true, nil
+	} else if assetAgreement.Recipient == "*" && assetLockVal.Locker != assetAgreement.Locker {
+		errorMsg := fmt.Sprintf("Asset of type %s and ID %s is not locked by %s", assetAgreement.Type, assetAgreement.Id, assetAgreement.Locker)
+		log.Error(errorMsg)
+		return false, errors.New(errorMsg)
+	} else if assetLockVal.Locker != assetAgreement.Locker || assetLockVal.Recipient != assetAgreement.Recipient {
+		errorMsg := fmt.Sprintf("Asset of type %s and ID %s is not locked by %s for %s", assetAgreement.Type, assetAgreement.Id, assetAgreement.Locker, assetAgreement.Recipient)
+		log.Error(errorMsg)
+		return false, errors.New(errorMsg)
+	}
```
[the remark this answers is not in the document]
knarayan: Yes @ermyas, the above if condition looks simplified. Updated.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,310 @@
...
+func (s *SmartContract) LockAsset(ctx contractapi.TransactionContextInterface, assetAgreementBytes string, lockInfoBytes string) error {
+
+	assetAgreement := &common.AssetExchangeAgreement{}
+	err := proto.Unmarshal([]byte(assetAgreementBytes), assetAgreement)
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+	//display the requested asset agreement
+	fmt.Printf("assetExchangeAgreement: %+v\n", assetAgreement)
```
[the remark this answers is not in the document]
knarayan: So specific reason @ermyas. Earlier I was only logging the errors, and writing the info messages to stdout. Updated to write the info messages to logger.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,310 @@
...
+
+	claimInfo := &common.AssetClaimHTLC{}
+	err = proto.Unmarshal([]byte(claimInfoBytes), claimInfo)
+	if err != nil {
+		//log.Error(err.Error())
+		return err
+		errorMsg := fmt.Sprintf("Unmarshal error: %s", err)
+		log.Error(errorMsg)
+		return errors.New(errorMsg)
```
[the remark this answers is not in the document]
knarayan: Good catch @ermyas. Updated.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	if assetLockVal.Locker != assetAgreement.Locker || assetLockVal.Recipient != assetAgreement.Recipient {
+		errorMsg := fmt.Sprintf("Cannot unlock asset of type %s and ID %s as it is locked by %s for %s", assetAgreement.Type, assetAgreement.Id, assetLockVal.Locker, assetLockVal.Recipient)
+		log.Error(errorMsg)
+		return errors.New(errorMsg)
+	}
+
+	// Check if expiry time is elapsed
+	currentTimeSecs := time.Now().UnixNano() / int64(time.Second)
```
ermyas, line 124: is there a reason we can't just use `time.Now().Unix()` instead? https://golang.org/pkg/time/#Time.Unix
_comment applies to the 5 other similar instances_

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	if uint64(currentTimeSecs) >= assetLockVal.ExpiryTimeSecs {
+		errorMsg := fmt.Sprintf("Can not claim asset of type %s and ID %s as the expiry time is already elapsed", assetAgreement.Type, assetAgreement.Id)
+		log.Error(errorMsg)
+		return errors.New(errorMsg)
+	}
+
+	err = ctx.GetStub().DelState(assetLockKey)
```
ermyas, line 296: How would a counter-party in the second network learn about the pre-image? Recall that we require the counter-party to be able to learn about the pre-image when the pre-image is first revealed on-chain in the second network so that it can then use it to claim its rightful asset on the first network (we can't assume or rely on off-chain communication here). If I am not mistaken the code above does not persist the pre-image, and in fact, removes the agreement at the end of a claim.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	"github.com/hyperledger/fabric-contract-api-go/contractapi"
+	"github.com/golang/protobuf/proto"
+	log "github.com/sirupsen/logrus"
+	"github.com/hyperledger-labs/weaver-dlt-interoperability/core/network/fabric-interop-cc/contracts/interop/protos-go/common"
+)
+
+type AssetLockValue struct {
+	Locker	string	`json:"locker"`
+	Recipient	string	`json:"recipient"`
+	Hash	string	`json:"hash"`
```
VRamakrishna, line 28: Perhaps align the type declarations vertically using tabs and whitespaces?

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	//display the requested asset agreement
+	log.Info(fmt.Printf("lockInfoHTLC: %+v\n", lockInfoHTLC))
+
+	if lockInfoHTLC.TimeSpec != common.AssetLockHTLC_EPOCH {
+		errorMsg := "only EPOCH time is supported at present"
+		log.Error(errorMsg)
+		return errors.New(errorMsg)
+	}
+
+	assetLockKey := assetAgreement.Type + ":" + assetAgreement.Id
```
VRamakrishna, line 59: I wonder if the locker also ought to be in the key? I know we are considering for now that an asset of a given type has a unique ID within a network, but perhaps that assumption may not always hold up?

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+		errorMessage := fmt.Sprintf("Failed to delete lock for asset of type %s and ID %s: %v", assetAgreement.Type, assetAgreement.Id, err)
+		log.Error(errorMessage)
+		return errors.New(errorMessage)
+	}
+
+	return nil
+}
+
+// IsAssetLocked cc is used to query the ledger and findout if an asset is locked or not
+func (s *SmartContract) IsAssetLocked(ctx contractapi.TransactionContextInterface, assetAgreementBytes string) (bool, error) {
```
VRamakrishna, line 142: I don't see any time checks in this function. I believe we'd determined in our design discussions that an asset's lock's status would be determined by time expiration and not solely through claims or reclaims.

I think we should add logic here whereby the function returns a false if the current time exceeds the expiry time associated with a lock. Since this is a query, we won't update or delete state, but leave it to the user to submit a Claim or an Unlock, as the case may be.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	if err != nil {
+		errorMsg := fmt.Sprintf("Base64 decode preimage error: %s", err)
+		log.Error(errorMsg)
+		return false, errors.New(errorMsg)
+	}
+
+	hasher := sha256.New()
+	hasher.Write([]byte(preimage))
+	shaHash := hasher.Sum(nil)
+	//shaHashBase64 := base64.URLEncoding.EncodeToString(shaHash)
```
VRamakrishna, line 211: Remove line if not needed?

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+
+	if isCorrectPreimage == false {
+		errorMsg := fmt.Sprintf("Can not claim asset of type %s and ID %s as the hash preimage is not matching", assetAgreement.Type, assetAgreement.Id)
+		log.Error(errorMsg)
+		return errors.New(errorMsg)
+	}
+
+	// Check if expiry time is elapsed
+	currentTimeSecs := time.Now().UnixNano() / int64(time.Second)
```
VRamakrishna, line 289: Change `UnixNano()` to `Unix()`.

Also, from an efficiency perspective, I'd move this check above (i.e., before) the hash preimage check. If the time's expired, we'll save some computation cycles that would otherwise be spent in hashing.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
+/*
+ * Copyright IBM Corp. All Rights Reserved.
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+// manage_assets is a chaincode that contains all the code related to asset management operations (e.g., Lock, Unlock, Claim)
+// and any related utility functions
+package main
```
VRamakrishna, line 9: I assume you will work on the list query functions in a later iteration?

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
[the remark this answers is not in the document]
VRamakrishna: Yes, I agree with having a single `getKey` function (or two, for the different categories). It's a typical pattern we use in a lot of chaincodes.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
VRamakrishna: Counterparty will learn about the preimage through events (not yet implemented) in the app layer. I don't see a need to persist it in the ledger. If push comes to shove, we can use a combination of history queries and block fetches to figure out what the preimage was, but it can be stored and used just in the app layer.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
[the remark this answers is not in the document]
knarayan: @ermyas and @VRamakrishna, Implemented the contractId and assetLockKey generation function. But I feel that it's use can be realized in the fungible asset scenarios fully.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+	assetLockKey := assetAgreement.Type + ":" + assetAgreement.Id
+
+	assetLockValBytes, err := ctx.GetStub().GetState(assetLockKey)
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+
+        if assetLockValBytes == nil {
+		errorMsg := fmt.Sprintf("No asset of type %s and ID %s is locked", assetAgreement.Type, assetAgreement.Id)
```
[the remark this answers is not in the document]
knarayan: Done.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,304 @@
...
+func (s *SmartContract) UnLockAsset(ctx contractapi.TransactionContextInterface, assetAgreementBytes string) error {
+
+	assetAgreement := &common.AssetExchangeAgreement{}
+	err := proto.Unmarshal([]byte(assetAgreementBytes), assetAgreement)
+	if err != nil {
+		log.Error(err.Error())
+		return err
+	}
+	//display the requested asset agreement
+	log.Info(fmt.Printf("assetExchangeAgreement: %+v\n", assetAgreement))
```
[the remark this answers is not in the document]
knarayan: I tried to use Infof() but then the messages are not getting printed to stdout. Hence I thought that I will retain this.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Done.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Yes, the preimage is already available on the ledger of the second network.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Done @VRamakrishna

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Done.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Done @VRamakrishna .

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -0,0 +1,321 @@
...
+	Locker		string	`json:"locker"`
+	Recipient	string	`json:"recipient"`
+	Hash		string	`json:"hash"`
+	ExpiryTimeSecs	uint64	`json:"expiryTimeSecs"`
+}
+
+// function to generate contract id as "SHA256" hash based on the asset lock key (which is combination of asset type and asset it)
+func generateAssetLockKeyAndContractId(assetAgreement *common.AssetExchangeAgreement) (string, string) {
+	hasher := sha256.New()
+	assetLockKey := assetAgreement.Type + assetAgreement.Id
```
VRamakrishna, line 36: As I mentioned earlier, it's probably useful to add the locker to the key as well, just to disambiguate future assets that may have the same type and id but belong to different owners.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
VRamakrishna: Where did we reach this conclusion? I still believe the locker ought to be part of the key (this will help with indexing when you implement list queries). In fact, the contract ID is also needed, I think. When you implement fungible assets, just using asset type and number of units won't suffice for the key. You will need the contract ID to generate a unique ledger key for the asset lock.

PR #47, opened 2021-05-09
Added SimpleAsset application chaincode, and support for asset exchange in fabric-cli.

sanvenDev:
Changes:
1. Added a sample application chaincode: simpleasset, to manage lifecycle of an asset.
2. Added two commands in fabric-cli to allow calling a demo asset exchange. (`fabric-cli asset exchange-all`, and `fabric-cli asset exchange-step`).

Please review and leave your comments if any changes required.


samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+type SmartContract struct {
+	contractapi.Contract
+}
+
+type Asset struct {
+	ID             string 			`json:"id"`
+	AssetType      string 			`json:"assettype"`
+	Owner          string 			`json:"owner"`
+	Issuer         string 			`json:"issuer"`
+	FaceValue 		 int    			`json:"facevalue"`
```
VRamakrishna, line 23: Can you fix the spacing so the different fields are aligned?

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	if !exists {
+		return fmt.Errorf("the asset %s does not exist", id)
+	}
+
+	return ctx.GetStub().DelState(id)
+}
+
+// AssetExists returns true when asset with given ID exists in world state
+func (s *SmartContract) isAssetExists(ctx contractapi.TransactionContextInterface, id string) (bool, error) {
```
VRamakrishna, line 115: Perhaps change the function name to `doesAssetExist` for grammatical correctness?

VRamakrishna: Also, for consistency, can we have all function names representing chaincode transactions begin with a capital letter? I checked other chaincode modules in the repository and that's the naming scheme we've been using.

sanvenDev: Okay, I too thought about `doesAssetExist`, but it seemed strange as a function name, what about just `AssetExists`?
`if (AssetExists)` would be better programmatically right?

samples/simpleasset/mocks/chaincodestub.go
```diff
@@ -0,0 +1,2878 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
VRamakrishna, line 1: This looks like a copy of the mock stub files in the Fabric Interop CC? We shouldn't have multiple copies of the same file in different places in a repo.

I'd suggest creating a makefile and adding copy instructions there, just like we do for the protos.

samples/simpleasset/mocks/statequeryiterator.go
```diff
@@ -0,0 +1,232 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
VRamakrishna, line 1: Also a copy. Same suggestion as above.

samples/simpleasset/mocks/transaction.go
```diff
@@ -0,0 +1,164 @@
+// Code generated by counterfeiter. DO NOT EDIT.
```
VRamakrishna, line 1: Also a copy. Same suggestion as above.

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/hyperledger/fabric-chaincode-go/shim"
+	"github.com/hyperledger/fabric-contract-api-go/contractapi"
+)
+
+// SmartContract provides functions for managing an Asset
+type SmartContract struct {
+	contractapi.Contract
+}
+
+type Asset struct {
```
VRamakrishna, line 18: It looks like this is loosely meant to represent a bond? Why not call it a `BondAsset` then?

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+	ID             string 			`json:"id"`
+	AssetType      string 			`json:"assettype"`
```
VRamakrishna: I believe this field exists to help us identify an instance of this asset during an exchange? If so, I think the "type" characteristic ought to exist outside of the asset itself. Let this chaincode just manage the asset without worrying about what needs to be done for atomic swaps. When we instrument/adapt the code for HTLC, we can add type definitions, but outside the specs of the assets themselves.

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+	Issuer         string 			`json:"issuer"`
+	FaceValue 		 int    			`json:"facevalue"`
+	MaturityDate 	 time.Time    `json:"maturitydate"`
+}
+
+
+// InitLedger adds a base set of assets to the ledger
+func (s *SmartContract) InitLedger(ctx contractapi.TransactionContextInterface) error {
+	assets := []Asset{
+		{ID: "a1", AssetType: "type1", Issuer: "Treasury" , Owner: "", FaceValue: 300,
```
VRamakrishna, line 31: As mentioned above, I'd get rid of the type field altogether.

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+	exists, err := s.isAssetExists(ctx, id)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return fmt.Errorf("the asset %s already exists", id)
+	}
+
+	asset := Asset{
+		ID:             id,
```
VRamakrishna: Align values vertically using whitespaces/tabs.

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+		if err != nil {
+			return nil, err
+		}
+		assets = append(assets, &asset)
+	}
+
+	return assets, nil
+}
+
+func main() {
```
VRamakrishna, line 211: This chaincode serves the purpose for a non-fungible asset. But we'd like to support one kind of fungible asset as well, say a `Token` or a `DigitalCash` (to be more explicit). Do you want to add code for that in this PR or in another one?

samples/simpleasset/simpleasset.go
VRamakrishna: Sure, `AssetExists` is fine.

samples/simpleasset/mocks/chaincodestub.go
sanvenDev: Okay, actually it was a copy from fabric-samples repo, I didn't realise it was present in fabric-interop-cc as well.
It was only for the purpose of testing the chaincode using `go test` command. 
Do you think using makefile for that is better? If yes I can do that.

samples/simpleasset/simpleasset.go
```diff
@@ -0,0 +1,238 @@
...
+func (s *SmartContract) isAssetExists(ctx contractapi.TransactionContextInterface, id string) (bool, error) {
+	assetJSON, err := ctx.GetStub().GetState(id)
+	if err != nil {
+		return false, fmt.Errorf("failed to read from world state: %v", err)
+	}
+
+	return assetJSON != nil, nil
+}
+// AssetExists returns true when asset with given ID exists in world state
+func (s *SmartContract) isAssetReleased(ctx contractapi.TransactionContextInterface, id string) (bool, error) {
```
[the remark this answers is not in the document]
sanvenDev: No Rama, this AssetReleased means in the context of expiry of maturity date and then the asset is released (i.e. no longer an active asset, this is what I understood from conversation with Krishna).

samples/simpleasset/simpleasset.go
sanvenDev: I'll add that code in next PR. Do you suggest a separate chaincode for that? because the functions might behave differently for a fungible asset, so I was thinking it to be a different chaincode.

samples/simpleasset/simpleasset.go
sanvenDev: I and Krishna think asset type can be something independent of asset exchange right? What do you say?

sanvenDev:
Creating separate pull request for chaincode, and then fabric-cli later on after API is ready.

PR #49, opened 2021-05-12
Fungible and non-fungible asset functions using contractId

knarayan:
Implements the below non-fungible asset functions:
- IsAssetLockedUsingContractId
- UnLockAssetUsingContractId
- ClaimAssetUsingContractId

Implements the below fungible asset functions:
- LockFungibleAsset
- IsFungibleAssetLocked
- UnLockFungibleAsset
- ClaimFungibleAsset


knarayan:
Update the `base class` functions for the `fungible asset` operations (lock, claim, unlock, is-asset-locked) to work using the `contractId` api.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -30,15 +32,57 @@ type AssetLockValue struct {
 	ExpiryTimeSecs	uint64	`json:"expiryTimeSecs"`
 }
 
-// function to generate contract id as "SHA256" hash based on the asset lock key (which is combination of asset type and asset it)
-func generateAssetLockKeyAndContractId(assetAgreement *common.AssetExchangeAgreement) (string, string) {
+// Object used in the map, contractId --> <asset-type, num-units, locker, ...> (for fungible assets)
+type FungibleAssetLockValue struct {
```
VRamakrishna, line 36: Don't you need a `ContractId` field here just as in `AssetLockValue`? If your data map separates the contract Id from the lock information and uses the former purely as a primary key to look up the contract details, then I would recommend you remove it from `AssetLockValue` too, and use suitable RW sets that enable an O(1) lookup based on this field for either category of asset.

For fungible assets, you can index your storage using <contract id>--><contract lock info>
For non-fungible assets, you can index your storage using <contract id>--><asset type, asset id>--><contract lock info>.
This produces some consistency between the two categories and also allows fast lookup based on <contract id> as well as <asset type, asset id> in the case of non-fungible assets.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -30,15 +32,57 @@ type AssetLockValue struct {
...
+	shaHashBase64 := base64.StdEncoding.EncodeToString(shaHash)
+	return shaHashBase64
 }
+
+/*
+ * Function to generate asset-lock key (which is combination of asset-type and asset-id)
+ * and contract-id (which is a hash on asset-lock key) for the non-fungible asset locking on the ledger
+ */
+func generateAssetLockKeyAndContractId(assetAgreement *common.AssetExchangeAgreement) (string, string) {
+	assetLockKey := assetKeyPrefix + assetAgreement.Type + assetKeyDelimiter + assetAgreement.Id
```
VRamakrishna, line 65: Upon further reflection, you can use the composite key creation function here rather than explicitly specify prefixes and delimiters. This will also allow us to query on the basis of just asset type, should the need arise.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -30,15 +32,57 @@ type AssetLockValue struct {
...
+}
+
+/*
+ * Function to generate contract-id for fungible asset-locking on the ledger (which is
+ * a hash on the attributes of the fungible asset exchange agreement)
+ */
+func generateFungibleAssetLockContractId(assetAgreement *common.FungibleAssetExchangeAgreement) string {
+	// There can be multiple fungible asset lock agreement between the same parties for the same <asset-type, quantity>.
+	// Use the current time to generate a different hash if the request gets repeated. It's the caller's responsibility
+	// to ensure that the request didn't get repeated uniterntionally.
```
VRamakrishna, line 77: Fix spelling of "unintentionally".

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -101,7 +145,7 @@ func (s *SmartContract) LockAsset(ctx contractapi.TransactionContextInterface, a
 		return "", errors.New(errorMsg)
 	}
 
-	err = ctx.GetStub().PutState(string(contractId), assetLockKeyBytes)
+	err = ctx.GetStub().PutState(contractIdPrefix + string(contractId), assetLockKeyBytes)
```
VRamakrishna, line 148: If you are not using the `contractId` field in this function for any other purpose, perhaps you can return the string with the prefix from the function that generates the contract ID itself?

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -338,3 +379,349 @@ func (s *SmartContract) ClaimAsset(ctx contractapi.TransactionContextInterface,
...
+		errorMessage := fmt.Sprintf("failed to delete the contractId %s as part of asset claim: %+v", contractId, err)
+		log.Error(errorMessage)
+		return errors.New(errorMessage)
+	}
+
+	return nil
+}
+
+// IsAssetLocked cc is used to query the ledger and find out if an asset is locked or not (this uses the contractId)
+func (s *SmartContract) IsAssetLockedUsingContractId(ctx contractapi.TransactionContextInterface, contractId string) (bool, error) {
```
VRamakrishna, line 516: The function name reads weird, as if we are asking whether the asset is locked using a contract Id. But I'm not sure what a suitable name would be.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -32,37 +34,17 @@ func (am *AssetManagement) Configure(interopChaincodeId string) {
     am.interopChaincodeId = interopChaincodeId
 }
 
+// function to generate a "SHA256" hash in base64 format for a given the preimage
```
VRamakrishna, line 37: Remove "the" before "preimage"

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -304,7 +286,10 @@ func (am *AssetManagement) IsFungibleAssetLocked(stub shim.ChaincodeStubInterfac
         log.Error(errorMsg)
         return false, errors.New(errorMsg)
     }
-    iccResp := stub.InvokeChaincode(am.interopChaincodeId, [][]byte{[]byte("IsFungibleAssetLocked"), []byte(assetAgreement.Type), []byte(strconv.FormatInt(int64(assetAgreement.NumUnits), 10)), []byte(assetAgreement.Recipient), []byte(assetAgreement.Locker)}, "")
+
+    contractId := generateSHA256HashInBase64Form(assetAgreement.Type + ":" + strconv.Itoa(int(assetAgreement.NumUnits)) +
+			":" + assetAgreement.Locker + ":" + assetAgreement.Recipient)
```
VRamakrishna, line 291: I don't think this is correct. The idea here was to have the `LockFungibleAsset` function return a unique contract Id that the caller could store and then use in functions like `IsFungibleAssetLocked`, `ClaimFungibleAsset`, and `UnlockFungibleAsset`. The above contract Id generation will produce collisions across multiple locks over similar quantities of an asset made by a given party for another.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -429,17 +414,15 @@ func (am *AssetManagement) ClaimFungibleAsset(stub shim.ChaincodeStubInterface,
...
-    }
     claimInfoBytes, err := proto.Marshal(claimInfoHTLC)
     if err != nil {
         log.Error(err.Error())
         return false, err
     }
-    iccResp := stub.InvokeChaincode(am.interopChaincodeId, [][]byte{[]byte("ClaimFungibleAsset"), assetAgreementBytes, claimInfoBytes}, "")
+    caller, _ := stub.GetCreator()
+    contractId := generateSHA256HashInBase64Form(assetAgreement.Type + ":" + strconv.Itoa(int(assetAgreement.NumUnits)) +
+			":" + assetAgreement.Locker + ":" + string(caller))
```
VRamakrishna, line 424: Same comment as above.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -510,7 +493,10 @@ func (am *AssetManagement) UnlockFungibleAsset(stub shim.ChaincodeStubInterface,
         log.Error(errorMsg)
         return false, errors.New(errorMsg)
     }
-    iccResp := stub.InvokeChaincode(am.interopChaincodeId, [][]byte{[]byte("UnlockFungibleAsset"), []byte(assetAgreement.Type), []byte(strconv.FormatInt(int64(assetAgreement.NumUnits), 10)), []byte(assetAgreement.Recipient)}, "")
+    caller , _ := stub.GetCreator()
+    contractId := generateSHA256HashInBase64Form(assetAgreement.Type + ":" + strconv.Itoa(int(assetAgreement.NumUnits)) +
+			":" + string(caller) + ":" + assetAgreement.Recipient)
```
VRamakrishna, line 498: Same comment as above.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -30,15 +32,57 @@ type AssetLockValue struct {
...
+ * a hash on the attributes of the fungible asset exchange agreement)
+ */
+func generateFungibleAssetLockContractId(assetAgreement *common.FungibleAssetExchangeAgreement) string {
+	// There can be multiple fungible asset lock agreement between the same parties for the same <asset-type, quantity>.
+	// Use the current time to generate a different hash if the request gets repeated. It's the caller's responsibility
+	// to ensure that the request didn't get repeated uniterntionally.
+	currentTimeSecs := uint64(time.Now().Unix())
+
+	preimage := assetAgreement.Type + strconv.Itoa(int(assetAgreement.NumUnits)) +
+			assetAgreement.Locker + assetAgreement.Recipient +  strconv.Itoa(int(currentTimeSecs))
```
VRamakrishna, line 81: This won't work, because this combination of parameter values can be repeated in multiple lock instances (e.g., Alice locks 20 ERC tokens for Bob in multiple instances). So it will end up producing the same hash in multiple instances.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna , I have removed `ContractId` field from `AssetLockValue` too as suggested. Yes, the indexing is now as per what you have summarized above and consistent.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: I assume that you are suggesting something like this https://stackoverflow.com/questions/52348514/how-to-make-composite-key-for-a-hash-map-in-golang to be used here too.
I want to figure this out later and include the code changes appropriately along with other code changes that I will be doing to address the query functions. Didn't do any changes now. Hope this is fine with you too @VRamakrishna .

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Good catch, thanks!

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna , the prefix here is internal to `interop-cc` and helps avoid conflicts with other types of keys on the ledger. Hence, I prefer not to expose to the user.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Changed the function name from `IsAssetLockedUsingContractId` to `IsAssetLockedQueryUsingContractId`.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
knarayan: Done!

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
knarayan: Made changes to address this behavior.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
knarayan: Made changes to address this behavior.

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
knarayan: Made changes to address this behavior.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna, The hash generated also depends on `currentTimeSecs`. Hence even if the `FungibleAssetExchangeAgreement` parameters are same across different requests, we get different hash values. Made no changes.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna , created composite function using the asset-type and asset-id attributes as suggested (https://pkg.go.dev/github.com/hyperledger/fabric-chaincode-go/shim#ChaincodeStub.CreateCompositeKey).

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna crated a function `generateContractIdMapKey` to return the key for the map using contractId.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: @VRamakrishna, replaced `currentTimeSecs` with the transactionId(https://pkg.go.dev/github.com/hyperledger/fabric-chaincode-go/shim#ChaincodeStub.GetTxID) to ensure uniqueness of the key across duplicate asset exchange agreements, and to maintain consistency of the write-set among multiple endorsements (`currentTimeSecs` might be different on different endorsing peers hence the write-sets from different endorsing peers may not match for the key `contractId` which is generated using `currentTimeSecs`).

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -338,3 +401,349 @@ func (s *SmartContract) ClaimAsset(ctx contractapi.TransactionContextInterface,
...
+		errorMsg := fmt.Sprintf("contractId %s is not associated with any currently locked asset", contractId)
+		log.Error(errorMsg)
+		return assetLockKey, assetLockVal, errors.New(errorMsg)
+	}
+
+	err = json.Unmarshal(assetLockValBytes, &assetLockVal)
+	if err != nil {
+		errorMsg := fmt.Sprintf("unmarshal error: %s", err)
+		log.Error(errorMsg)
+		return assetLockKey, assetLockVal, errors.New(errorMsg)
```
ermyas, line 439: This 3-line code snippet is repeated about 157 times (roughly 470 LoC) across `manage_assets.go` and `asset_locks.go` with minor variations. I think it clutters the code a bit. Given that all this code does is log the error before creating it, I think we can reduce the three lines to a single call to a helper function. The helper function (`logThenErrorf(...)` or something along those lines) could take a variadic argument of the formatted error string and its parameters, from which it can create an error message, log it and return an associated `error` instance. For places that already have an `error` object (e.g. line 120 `managed_assets.go`), we can have a function that takes the error, logs it, and returns it back.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -57,7 +106,7 @@ func (s *SmartContract) LockAsset(ctx contractapi.TransactionContextInterface, a
 		log.Error(err.Error())
 		return "", err
 	}
-	//display the requested asset agreement
+	//display the passed lock information
 	log.Info(fmt.Sprintf("lockInfoHTLC: %+v\n", lockInfoHTLC))
```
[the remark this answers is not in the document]
VRamakrishna: What we currently have is better than the earlier `log.Info(fmt.Printf(....))`, but I agree `log.Infof(....)` looks better.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -179,7 +237,11 @@ func (s *SmartContract) IsAssetLocked(ctx contractapi.TransactionContextInterfac
 	//display the requested asset agreement
 	log.Info(fmt.Sprintf("assetExchangeAgreement: %+v\n", assetAgreement))
 
-	assetLockKey, _ := generateAssetLockKeyAndContractId(assetAgreement)
+	assetLockKey, _, err := generateAssetLockKeyAndContractId(ctx, assetAgreement)
```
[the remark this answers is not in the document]
VRamakrishna: @ermyas `time.Now().Unix()` returns type `int64`, hence the cast is needed.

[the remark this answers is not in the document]
ermyas: @VRamakrishna I understand the cast when we are assigning the variable, what I am wondering is why we again cast it in if statement

VRamakrishna (approved):

PR #55, opened 2021-05-19
changes to receive serialized protos in base64 and validation on locker/recipient

knarayan:


core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -93,18 +93,107 @@ func generateFungibleAssetLockContractId(ctx contractapi.TransactionContextInter
...
+func validateAndSetPartiesOfAssetAgreement(ctx contractapi.TransactionContextInterface, assetAgreement *common.AssetExchangeAgreement) error {
+	txCreator, err := ctx.GetStub().GetCreator()
+	if err != nil {
+		return fmt.Errorf("unable to get the transaction creator information")
+	}
+	if len(assetAgreement.Locker) == 0 {
+		assetAgreement.Locker = string(txCreator)
+	} else if len(assetAgreement.Recipient) == 0 {
+		assetAgreement.Recipient = string(txCreator)
+	} else if assetAgreement.Locker != string(txCreator) && assetAgreement.Recipient != string(txCreator) {
```
VRamakrishna, line 163: Minor optimization: you can assign `string(txCreator)` to a variable and use that everywhere.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -93,18 +93,107 @@ func generateFungibleAssetLockContractId(ctx contractapi.TransactionContextInter
 	return contractId
 }
 
+/*
+ * Function to validate the locker in asset agreement.
+ * If locker is not set, it will be set to the caller.
+ * If the locker is set already, it ensures that the locker is same as the creator of the transaction.
+ */
+func validateAndSetLockerOfAssetAgreement(ctx contractapi.TransactionContextInterface, assetAgreement *common.AssetExchangeAgreement) error {
```
VRamakrishna, line 101: You can probably just pass `Locker` instead of the whole agreement here, so this function can be called by the locking functions that rely on contract IDs.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -93,18 +93,107 @@ func generateFungibleAssetLockContractId(ctx contractapi.TransactionContextInter
...
+
+	return nil
+}
+
+/*
+ * Function to validate the recipient in asset agreement.
+ * If recipient is not set, it will be set to the caller.
+ * If the recipeint is set already, it ensures that the recipient is same as the creator of the transaction.
+ */
+func validateAndSetRecipientOfAssetAgreement(ctx contractapi.TransactionContextInterface, assetAgreement *common.AssetExchangeAgreement) error {
```
VRamakrishna, line 139: You can probably just pass `Recipient` instead of the whole agreement here, so this function can be called by the locking functions that rely on contract IDs.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -655,6 +821,15 @@ func (s *SmartContract) IsFungibleAssetLocked(ctx contractapi.TransactionContext
 		return false, err
 	}
 
+	txCreator, err := ctx.GetStub().GetCreator()
+	if err != nil {
+		return false, fmt.Errorf("unable to get the transaction creator information: %+v", err)
+	}
+
+	if assetLockVal.Locker != string(txCreator) && assetLockVal.Recipient != string(txCreator) {
```
VRamakrishna, line 829: I don't think we need this check. I'd originally added it in the base class, but after a discussion with you and Ermyas, remember removing this logic. I think it's OK to return lock info even if the caller is not a party to the agreement.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Done @VRamakrishna .

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: This is not possible since we are also setting `Locker` in case it was not set earlier.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: This is not possible since we are also setting `Recipient` in case it was not set earlier.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Now removed this check in all the query functions @VRamakrishna .

sanvenDev (approved):

PR #63, opened 2021-05-29
Check asset ownership in ReadAsset of bondasset app cc

knarayan:


samples/simpleasset/assetmgmt.go
```diff
@@ -65,15 +65,19 @@ func getLockInfoProtoBytesHTLC(hashBase64 []byte, expiryTimeSecs uint64) ([]byte
 }
 
 // asset specific checks (ideally an asset in a different application might implment checks specific to that asset)
```
VRamakrishna, line 67: `implment` --> `implement`

samples/simpleasset/bondasset.go
```diff
@@ -96,6 +97,16 @@ func (s *SmartContract) ReadAsset(ctx contractapi.TransactionContextInterface, a
 	if err != nil {
 		return nil, err
 	}
+	// In the update onwer context, the TxCreator will be the newOwner. Hence don't check if the TxCreator is the current owner.
```
VRamakrishna, line 100: `onwer` --> `owner`

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
```diff
@@ -136,7 +136,7 @@ func (am *AssetManagement) LockAsset(stub shim.ChaincodeStubInterface, assetAgre
     iccResp := stub.InvokeChaincode(am.interopChaincodeId, [][]byte{[]byte("LockAsset"), []byte(assetAgreementBytes64), []byte(lockInfoBytes64)}, "")
     fmt.Printf("Response from Interop CC: %+v\n", iccResp)
     if iccResp.GetStatus() != shim.OK {
-        return "", logThenErrorf(string(iccResp.GetPayload()))
+        return "", logThenErrorf(string(iccResp.Message))
```
VRamakrishna, line 139: Returning payload instead of message is fine, but I wonder if the message may also contain something useful if the invocation didn't work (sometimes the failure occurs outside the chaincode function). Maybe return both in a JSON? Also, I think the message can be obtained using GetMessage().

samples/simpleasset/assetmgmt.go
knarayan: Done @VRamakrishna !

samples/simpleasset/bondasset.go
knarayan: Done!

core/network/fabric-interop-cc/interfaces/asset-mgmt/asset_locks.go
knarayan: I have updated `iccResp.Message` with `iccResp.GetMessage()`. The Payload() can be very long as well, and I guess logging that is good enough so that the chaincode container log can be inspected in case of any errors.

VRamakrishna (approved):

PR #74, opened 2021-06-16
code to query simple CorDapp using linearId

knarayan:


tests/network-setups/shared/corda/makefile
```diff
@@ -19,6 +19,6 @@ stop:
 	./scripts/stop-nodes.sh
 
 .PHONY: clean
-clean:	stop
+clean-local:	stop
```
sanvenDev, line 22: Clean for local build or artifactory(remote) build, is same. So I think `clean` would be better target name.

samples/corda-simple-application/workflows-kotlin/src/main/kotlin/com/example/flow/SimpleFlow.kt
```diff
@@ -237,20 +238,43 @@ class DeleteState(val key: String) : FlowLogic<SignedTransaction>() {
  * @property key the key for the [SimpleState] to be retrieved.
  */
 @StartableByRPC
-class GetStateByKey(val key: String) : FlowLogic<ByteArray>() {
+class GetStateByKey(val key: String) : FlowLogic<List<SimpleState>>() {
```
sanvenDev, line 241: This should return ByteArray only, as this function is called by remote network as well.
If you want to convert ByteArray into SimpleState, I suggest doing that in corda-client.

tests/network-setups/shared/corda/makefile
knarayan: @sanvenDev I reverted this to `clean` and changed in the `Getting Started` page from `clean-local` to `clean`. I wanted to ensure that it will be the same target name in both the `makefile` here and in the `Getting Started` page.

samples/corda-simple-application/workflows-kotlin/src/main/kotlin/com/example/flow/SimpleFlow.kt
knarayan: Ok @sanvenDev , I have retained this as `ByteArray` and converted into `String` in the `corda-client`.

sanvenDev (approved):

PR #83, opened 2021-06-21
Adding support for Corda-Network2 data transfer.

sanvenDev:
1. Added certs for network2 in Corda network, and modified `Corda_Relay.toml` config file.
2. Modified the way Certs are added in Corda network, so that its no longer hard coded and is configurable without need to re-deploy network.
3. Added commands like `configure-create-all`, `configure-all`, `configure-data`, and `configure-network`, which works similar to fabric-cli commands, except that `configure-all <list of networks separated by comma>` to add all these networks' credentials in Corda-network.


sanvenDev:
Closing to fix DCO issue first.

PR #86, opened 2021-06-21
using generic lockInfo and claimInfo structs in interop-cc

knarayan:
Signed-off-by: Krishnasuri Narayanam <knaraya3@in.ibm.com>

### DIFF
--- a/core/network/fabric-interop-cc/contracts/interop/manage_assets.go
+++ b/core/network/fabric-interop-cc/contracts/interop/manage_assets.go
@@ -209,6 +209,8 @@ func (s *SmartContract) LockAsset(ctx contractapi.TransactionContextInterface, a
 		if err != nil {
 			return "", logThenErrorf(err.Error())
 		}
+	} else {
+		return "", logThenErrorf("lock mechanism is not supported")
 	}
 	//display the passed lock information
 	log.Infof("lockInfo: %+v\n", lockInfoHTLC)
@@ -434,6 +436,8 @@ func (s *SmartContract) ClaimAsset(ctx contractapi.TransactionContextInterface, 
 		if err != nil {
 			return logThenErrorf(err.Error())
 		}
+	} else {
+		return "", logThenErrorf("lock mechanism is not supported")
 	}
 
 	// display the claim information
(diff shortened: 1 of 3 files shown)


core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -198,13 +198,20 @@ func (s *SmartContract) LockAsset(ctx contractapi.TransactionContextInterface, a
...
 	if err != nil {
 		return "", logThenErrorf(err.Error())
 	}
+	lockInfoHTLC := &common.AssetLockHTLC{}
+        if lockInfo.LockMechanism == common.LockMechanism_HTLC {
+		err = proto.Unmarshal(lockInfo.LockInfo, lockInfoHTLC)
+		if err != nil {
+			return "", logThenErrorf(err.Error())
+		}
+	}
```
sanvenDev, line 212: Do you think adding an else statement, where we can return some error like "Lock Mechanism not supported", just for safety? I know it is going to be always set a valid value in interop-sdk, but I still I think its a good idea to do that.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
```diff
@@ -415,14 +422,22 @@ func (s *SmartContract) ClaimAsset(ctx contractapi.TransactionContextInterface,
...
 		return logThenErrorf("unmarshal error: %s", err)
 	}
 
+	claimInfoHTLC := &common.AssetClaimHTLC{}
+	if claimInfo.LockMechanism == common.LockMechanism_HTLC {
+		err = proto.Unmarshal(claimInfo.ClaimInfo, claimInfoHTLC)
+		if err != nil {
+			return logThenErrorf(err.Error())
+		}
+	}
```
sanvenDev, line 437: same comment as above for LockAsset.

core/network/fabric-interop-cc/contracts/interop/manage_assets.go
knarayan: Yes, it is good to have it here also @sanvenDev . Added now.

sanvenDev (approved):

PR #117, opened 2021-07-21
Clean Up following use of modules, fix Asset-Mgmt module

sanvenDev:
1. Deleted `protos-go` and `protos-js` directories.
2. Deleted scripts in various places which is used to copy protos and build manually, (since this is being done in common folder, use that if you want to build locally).
3. Deleted commented codes from various makefiles and scripts (If want to use local protos, point to folders in `/common`)
4. Use interopcc `v1.2.1` in testnet.
5. Added mocks folder to `asset-mgmt`, as otherwise using it as go module used to throw error that mocks not found.
6. Simpleasset modified to import `asset-mgmt` as module. To test it, manually copy folder `core/network/fabric-interop-cc/interfaces/asset-mgmt` from my fork to your path: `$GOPATH/pkg/mod/github.com/hyperledger-labs/weaver-dlt-interoperability/core/network/fabric-interop-cc/interfaces/asset-mgmt@v1.2.1`.

Tested all 6 Interop calls, and local unit tests, and end to end asset-exchange after these changes, and it works fine for me.

Update:
1.  Included GRPC protos in protos-go compilation. 
2. Fixed `go_packages` in protos.
3. Using scripts to fetch corda interop modules instead of using build.gradle (apparently this seems to be better option)
4. Docs: removed all local build steps, fixed typos.
5. Removed unused dependencies from `interoperation-node-sdk` and `fabric-driver`


tests/network-setups/fabric/dev/scripts/setupAppCC.sh
```diff
@@ -13,8 +13,8 @@ if [ -d "${CHAINCODE_PATH}/${CC_NAME}" ]; then
 fi
 cp -r ${APP_CC_PATH} ${CHAINCODE_PATH}/
 
-if [ "${CC_NAME}" = "simpleasset" ]; then
-  cp -r ${INTEROPCC_CHAINCODE_PATH}/fabric-interop-cc/interfaces/asset-mgmt ${CHAINCODE_PATH}/${CC_NAME}/
-fi
+# if [ "${CC_NAME}" = "simpleasset" ]; then
+#   cp -r ${INTEROPCC_CHAINCODE_PATH}/fabric-interop-cc/interfaces/asset-mgmt ${CHAINCODE_PATH}/${CC_NAME}/
+# fi
```
VRamakrishna, line 18: Can we just delete these lines? Any reason to retain them?

tests/network-setups/fabric/dev/scripts/setupCC.sh
```diff
@@ -5,7 +5,7 @@ CHAINCODE_PATH=$PWD/../shared/chaincode
 
 # interop cc module
 INTEROPCC_MOD=github.com/hyperledger-labs/weaver-dlt-interoperability/core/network/fabric-interop-cc/contracts/interop
-INTEROP_VERSION=
+INTEROP_VERSION=@v1.2.1
```
VRamakrishna, line 8: Do we need the '@v' here?

VRamakrishna: Don't need this.

tests/network-setups/fabric/dev/scripts/setupAppCC.sh
sanvenDev: Deleted.

VRamakrishna (approved):

PR #124, opened 2021-08-09
Downgrade fabric-protos-go version in weaver-protos-go

sanvenDev:
Signed-off-by: sandeep.nishad1 <sandeep.nishad1@ibm.com>


VRamakrishna:
This change is required because the Go SDK we are building depends on `fabric-sdk-go`, which itself depends on an older version of `fabric-protos-go`. Without this change, we see build errors as `weaver-protos-go` and and `fabric-sdk-go` depend on mutually incompatible versions of `fabric-protos-go`.

VRamakrishna (approved):

PR #129, opened 2021-08-12
Abstract out signing to an interface

yacovm:
Signed-off-by: Yacov Manevich <yacovm@il.ibm.com>


petermetz:
This is also something that different projects could collaborate on! cc: @hartm

PR #149, opened 2021-08-27
Github Workflows for Unit and Integration Tests; Added local-docker page

sanvenDev:
1. Added workflow for all Go unit tests.
2. Added workflow for node sdk unit tests.
3. Added workflow for corda interop app unit tests.
4. Added workflow for Locally built weaver components (host deployment): end-to-end data transfer test.
5. Added workflow for Imported Dockerized Weaver Components: end-to-end data transfer test.
6. Added workflow for end-to-end asset-exchange (fabric-fabric) test.

Fix:
Fabric driver build error fix.

Update:
1. Modified Dockerfiles and makefiles to build docker images locally.
2. Added docs page for `setup-local-docker`.


.github/workflows/data-transfer.yml
```diff
@@ -0,0 +1,479 @@
...
+          echo "username=${GITHUB_ACTOR}" >> github.properties
+          echo "password=${{ secrets.GITHUB_TOKEN }}" >> github.properties
+          echo "url=https://maven.pkg.github.com/${GITHUB_ACTOR}/weaver-dlt-interoperability" >> github.properties
+          cat github.properties
+        working-directory: tests/network-setups/corda
+        
+      - name: Start Corda Network
+        run: |
+          sed -i "/docker logs corda_partya_1 -f/"' s/^/#/' "scripts/start-nodes.sh"
+          make start &> corda-net.out &
```
VRamakrishna, line 57: Add `2>&1` so stderr is also directed to `corda-net.out`?

.github/workflows/data-transfer.yml
```diff
@@ -0,0 +1,479 @@
...
+        
+      # FABRIC CLI
+      - name: Setup Fabric CLI .npmrc
+        run: |
+          cp .npmrc.template .npmrc
+          sed -i "s/<personal-access-token>/${{ secrets.GITHUB_TOKEN }}/g" .npmrc
+          cat .npmrc
+        working-directory: samples/fabric/fabric-cli
+      - name: Build Fabric CLI
+        run: npm install
```
VRamakrishna, line 115: Use `make build` here, which is what we mention in the docs too. You'll have to install `yarn` (using npm) in the prerequisites for this.

.github/workflows/data-transfer.yml
```diff
@@ -0,0 +1,479 @@
...
+          
+      - name: Use Protoc 3.15
+        run: |
+          curl -LO https://github.com/protocolbuffers/protobuf/releases/download/v3.15.6/protoc-3.15.6-linux-x86_64.zip
+          unzip protoc-3.15.6-linux-x86_64.zip -d protoc
+          go get -u google.golang.org/protobuf/cmd/protoc-gen-go
+          go get -u google.golang.org/grpc/cmd/protoc-gen-go-grpc
+          
+    
+      # PROTOS
```
VRamakrishna, line 273: Also build the Go proto library. I know the synthesized files are checked into the repo but the purpose of running a fully local build is that we can test after making any changes to the schemas in `common/protos` too.

.github/workflows/go.yml
```diff
@@ -0,0 +1,138 @@
...
+      working-directory: samples/fabric/simpleasset
+
+    - name: Build
+      run: go build -v ./...
+      working-directory: samples/fabric/simpleasset
+      
+    - name: Test
+      run: go test -v ./...
+      working-directory: samples/fabric/simpleasset
+
```
VRamakrishna, line 138: No test for `simpleassetandinterop`?

.github/workflows/go.yml
```diff
@@ -0,0 +1,138 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.16
+        
+    - name: Local Setup
```
VRamakrishna, line 38: We are using the imported protos here, right? Not locally built ones?

samples/fabric/fabric-cli/scripts/ci-test.sh
```diff
@@ -0,0 +1,40 @@
+COUNT=0
```
VRamakrishna, line 1: Where is this script getting called from? And should we add a note in the Fabric-CLI readme about configuring and running this?

.github/workflows/data-transfer.yml
sanvenDev: `&>` also directs everything, even the stderr.

sanvenDev: I just found out in bash both serves the same purpose, and since this testing environment is running in ubuntu which has bash, so its fine?

samples/fabric/fabric-cli/scripts/ci-test.sh
sanvenDev: Oh sorry I didn't notice, this was a dummy script, wasn't supposed to be checked in (probably isn't debugged as well). I'll delete this file.

.github/workflows/data-transfer.yml
sanvenDev: Changing it to `make build`. I can add `yarn` but I'm thinking `yarn link` doesn't do anything wrt build or running the fabric-cli. `yarn link` is only when you want to import a node js package in another node js package.

sanvenDev: Changing it to `make build`. Doesn't `npm install` should install `yarn` also, otherwise I think docs needs to be updated as well?

Also I'm thinking `yarn link` doesn't do anything wrt build or running the fabric-cli. `yarn link` is only when you want to import a node js package in another node js package.
I know Readme suggests running `yarn link` and then running `fabric-cli`, but I don't think `yarn link` does anything.
Shall we remove it from `make build`?

.github/workflows/go.yml
sanvenDev: The place where you commented, here we are using locally built ones.
There's another job above this, which uses imported ones.

.github/workflows/go.yml
sanvenDev: Added.

.github/workflows/go.yml
sanvenDev: Actually I'm assuming the person updating protos, should always rebuild protos-go before checking in, and so that, that's always upto date whatever is in `common/protos-go`.

.github/workflows/data-transfer.yml
sanvenDev: Added build for protos-go

.github/workflows/data-transfer.yml
VRamakrishna: OK

.github/workflows/data-transfer.yml
VRamakrishna: The prerequisites section explicitly asks the user to install Yarn. That instruction has been there from the beginning. Do we need anything else?

VRamakrishna: Yeah, `yarn link` allows the `fabric-cli` to be available as a global executable. But you are right, we are only using it from the `bin` folder.

You can remove `yarn` if you want. Apart from the makefile and Readme, you'll also need to update all the "Setup...." pages in the docs and remove it from the prerequisites sections.

VRamakrishna (approved):

PR #152, opened 2021-09-01
Fabric CLI changes for asset-exchange using Go

knarayan:
This PR includes:

- user registration/enrollment using go-sdk
- 'fabric-cli user add ' command in go-cli
- Minor typos in docs and in fabric-cli for node


samples/fabric/fabric-cli/src/helpers/fabric-functions.ts
```diff
@@ -79,13 +79,15 @@ const walletSetup = async (
       logger.error(`Identity ${userName} does not exist. Please add user in the network.\n`)
       return
```
sanvenDev, line 80: Can we do one thing here? if register is false, still try the enrollment (in case user is registered but wallet id file is deleted, it can fetch the id, that's what my intention was initially to do). If enrollment fails, possibly user is not even registered then log this error `Identity ${userName} does not exist. Please add user in the network.\n`.

knarayan: @sanvenDev as discussed, the command ```$ ./bin/fabric-cli user add --target-network=network1 --id=alice --secret=secrettext``` can be used to recover the wallet for a user who is already registered and wallet is lost.

sanvenDev (dismissed):

sanvenDev (approved):

PR #154, opened 2021-09-06
Fabric Interoperation Chaincode Refactoring and Augmentation for Access Controls, and Logging Improvements

VRamakrishna:
- Separated Go mock testing functions into a separate `testutils` library (Go package).
- Added access control logic to Fabric Interop CC asset management module to allow invocation requests only from other application chaincodes; also associated HTLC units with a chaincode ID for isolation of locks within chaincode namespaces.
- Fixed logging behavior in relay and in the Corda Interop App to print complex payloads as Base64 strings rather than as byte arrays.
- Added support in Fabric CLI to run chaincodes other than `simplestate` for data sharing.
- Updated testnets for compatibility with the above changes and ran end-to-end tests.


tests/network-setups/fabric/dev/scripts/setupAppCC.sh
```diff
@@ -1,8 +1,7 @@
 CC_NAME=${1:-"simplestate"}
 
-INTEROPCC_CHAINCODE_PATH=$PWD/../../../../core/network
 CHAINCODE_PATH=$PWD/../shared/chaincode
-APP_CC_PATH=$PWD/../../../../samples/fabric/${CC_NAME}
+APP_CC_PATH=$PWD/../../../../samples/fabric/${CC_NAME}$2
```
sanvenDev, line 4: What is this `$2` for?

tests/network-setups/fabric/dev/scripts/setupAppCCLocal.sh
```diff
@@ -0,0 +1,24 @@
...
+
+echo "Copying ${CC_NAME}..."
+
+##################### Copying app chaincode ###############
+if [ -d "${CHAINCODE_PATH}/${CC_NAME}" ]; then
+    echo "Deleting existing ${CC_NAME} chaincode folder to copy the latest..."
+    rm -rf ${CHAINCODE_PATH}/${CC_NAME}
+fi
+cp -r ${APP_CC_PATH} ${CHAINCODE_PATH}/${CC_NAME}
+cp -r ${INTEROPCC_CHAINCODE_PATH}/fabric-interop-cc/libs/utils ${CHAINCODE_PATH}/${CC_NAME}/
```
sanvenDev, line 15: Is it really required to copy test related things into the chaincode? I think for deployment test related files were not required if `go.mod` and `go.sum` are already generated.

tests/network-setups/fabric/dev/scripts/setupCCLocal.sh
```diff
@@ -19,9 +19,16 @@ cp -r ${INTEROP_CC_PATH} .
 cd fabric-interop-cc
 cp -r $PWD/../../../../../../common/protos-go contracts/interop/
 cp -r $PWD/../../../../../../core/network/fabric-interop-cc/libs/assetexchange contracts/interop/
+cp -r $PWD/../../../../../../core/network/fabric-interop-cc/libs/utils contracts/interop/
+cp -r $PWD/../../../../../../core/network/fabric-interop-cc/libs/testutils contracts/interop/
```
sanvenDev, line 23: Same comment as above. if it requires `go.sum` may be check-in `go.sum.local` as well? Then these test utils won't be required?

tests/network-setups/fabric/dev/scripts/setupAppCC.sh
VRamakrishna: This was an option I'd added earlier and passed in the Makefile, but then thought better of it and removed it. But I forgot to remove the `$2` from the scripts. Will do that.

tests/network-setups/fabric/dev/scripts/setupAppCCLocal.sh
VRamakrishna: `utils` isn't related only to tests. It's imported in the source files.

core/network/fabric-interop-cc/contracts/interop/handle_external_request.go
```diff
@@ -28,6 +29,16 @@ import (
 // 3. Checks the access control policy for the requester and view address is met
 // 4. Calls application chaincode
 func (s *SmartContract) HandleExternalRequest(ctx contractapi.TransactionContextInterface, b64QueryBytes string) (string, error) {
+	if !s.testMode {
+		relayAccessCheck, err := wutils.CheckAccessIfRelayClient(ctx.GetStub())
```
sanvenDev, line 33: Shouldn't this be just `IsClientRelay`?

VRamakrishna: Yes. Fixed.

sanvenDev (approved):

PR #173, opened 2021-10-11
fix(build): relay build.rs relative paths for .protos

petermetz:
Signed-off-by: Peter Somogyvari <peter.somogyvari@accenture.com>


sanvenDev:
@petermetz There is a makefile in `/core/relay` directory, where there is a `build` target, that builds the `relay`.
It involves copying those protos to `core/relay` and then running `cargo build`.

The reason we do this is, during build of docker image for relay, we don't want to give whole weaver repo as context directory, which slows down the build.

So current path of protos in that file was intentional.

PR #180, opened 2021-10-12
Fabric-Fabric Asset Transfer Feature and Demonstration

VRamakrishna:
Fabric CLI Node refactoring and augmentation.
Documentation updates.
Fabric Interoperation and library chaincode upgrades.

### DIFF
--- a/.github/workflows/data-transfer.yml
+++ b/.github/workflows/data-transfer.yml
@@ -144,6 +144,7 @@ jobs:
         run: |
           echo ${GITHUB_WORKSPACE}
           cp .env.template .env
+          sed -i "s/CHAINCODE_PATH=.*/CHAINCODE_PATH=\.\/chaincode\.json/g" .env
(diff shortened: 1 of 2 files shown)


samples/fabric/fabric-cli/scripts/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "fabric-cli-scripts",
```
sanvenDev, line 2: Is this required?

samples/fabric/fabric-cli/scripts/getTokenBalance.sh
```diff
@@ -0,0 +1,7 @@
+cd scripts
+
+USER_CERT_BASE64=$(node getAssetTransferViewAddress.js getusercert $1 $2)
+
+cd ..
+
+./bin/fabric-cli chaincode query mychannel simpleassettransfer GetBalance '["token1","'$USER_CERT_BASE64'"]' --local-network=$1
```
sanvenDev, line 7: Is there a reason to do this way? In other `simpleasset`, we assume caller is the one who is querying his own balance (not allowing others to query balance of others).
And to set the caller user, in `query`/`invoke` you need pass extra argument `--user=<user-name>`.

PR #186, opened 2021-11-11
Besu-cli 

dhinakaran2705:
Base commit for besu-cli

Includes source file for all the commands

Readme to be improved


samples/besu/besu-cli/src/commands/generate.ts
```diff
@@ -0,0 +1,23 @@
+import { GluegunToolbox } from 'gluegun'
+
+module.exports = {
+  name: 'generate',
```
sanvenDev, line 4: Is this required?

samples/besu/besu-cli/src/commands/asset/issue.ts
```diff
@@ -0,0 +1,70 @@
...
+				`besu-cli asset issue --network=network1 --contract=path/to/contracts/tokenContract.json --account=1 --amount=10`,
+				'besu-cli asset issue --network=<network1|network2> --contract=<path-to-tokencontract-json>--account=<account-index> --amount=<issue-ammount>',
+				[
+					{
+						name: '--network',
+						description:
+							'network for command. <network1|network2>'
+					},
+					{
+						name: '--contract',
```
sanvenDev, line 27: Can we get this from `networkConfig` instead of passing it here? We are setting the contract there also.
Just to avoid this parameter.

samples/besu/besu-cli/src/commands/asset/lock.ts
```diff
@@ -0,0 +1,140 @@
...
+				`besu-cli asset lock --network=network1 --token_contract='path/to/tokenContract.json' --interop_contract='path/to/interopcontract.json' --sender_account=1 --recipient_account=2 --amount=5 --timeout=10`,
+				'besu-cli asset lock --network=<network1|network2> --token_contract=<path-to-tokenContract.json> --interop_contract=<path-to-interopContract.json> --sender_account=<1|2> --recipient_account=<2|1> --amount=<lock-amount> --timeout=<lock-duration-seconds> --hash=<hashLock-optional-parameter>',
+				[
+					{
+						name: '--network',
+						description:
+							'network for command. <network1|network2>'
+					},
+					{
+						name: '--token_contract',
```
sanvenDev, line 28: Do we need this, given that we are setting it in `config.json`.

samples/besu/besu-cli/src/commands/asset/lock.ts
```diff
@@ -0,0 +1,140 @@
...
+						description:
+							'Path to the json file corresponding to the token contract compiled with Truffle.'
+					},
+					{
+						name: '--interop_contract',
```
sanvenDev, line 33: Do we need this, given that we are setting it in `config.json`.

samples/besu/besu-cli/src/commands/asset/lock.ts
```diff
@@ -0,0 +1,140 @@
...
+						description:
+							'Path to the json file corresponding to the interop contract compiled with Truffle.'
+					},
+					{
+						name: '--sender_account',
```
sanvenDev, line 38: Do we need this, given that we are setting it in `config.json`.

samples/besu/besu-cli/src/commands/asset/lock.ts
```diff
@@ -0,0 +1,140 @@
...
+						description:
+							'The index of the account of the sender/owner of the asset from the list obtained through web3.eth.getAccounts(). For example, we can set Alice as accounts[1] and hence value of this parameter for Alice can be 1.'
+					},
+					{
+						name: '--recipient_account',
```
sanvenDev, line 43: Do we need this, given that we are setting it in `config.json`.

samples/besu/besu-cli/src/commands/asset/lock.ts
sanvenDev: Alternatively We can remove sender and receiver from `config.json`.

samples/besu/besu-cli/src/commands/asset/lock.ts
sanvenDev: Alternatively We can remove sender and receiver from `config.json`.

samples/besu/besu-cli/src/commands/generate.ts
dhinakaran2705: It looks like a generated file.

samples/besu/besu-cli/src/commands/asset/lock.ts
dhinakaran2705: I thought the accounts can be taken from the config for the exchange command and that other commands can take them via command line. What do you think?

If you don't favor that, I will go with taking from config.json for all the commands.

samples/besu/besu-cli/src/commands/asset/lock.ts
dhinakaran2705: Covered above.

sanvenDev (approved):
If possible fix the log of `is-locked` command. Approving it anyways, because regular asset exchange commands are working.

PR #211, opened 2022-03-28
Bug Fixes and Extra Log Messages

sanvenDev:
1. Bump Version for Relay and other fabric components, (after TLS addition we never did that).
2. Bug fix: Corda testnet (with house token app) startup fail with github packages.
3. Extra Log Messages in Fabric interop cc for access control and proof verification.
4. Return Hash from createHTLC function in fabric-node-sdk.


sdks/fabric/interoperation-node-sdk/package-local.json
```diff
@@ -6,7 +6,7 @@
     "blockchain",
     "interoperability"
   ],
-  "version": "1.2.1",
+  "version": "1.2.5",
```
VRamakrishna, line 9: Why can't we tag this as `1.2.4`?

VRamakrishna (approved):

PR #243, opened 2022-04-20
Minor Fixes in workflows and dependabot updates

sanvenDev:
1. Added Dependabot updates in this PR.
2. Fixed Deploy Docs workflow.
3. Added workflow to publish corda packages.
4. Added another job in fabric workflow tests, to test interopcc changes without publishing.
5. Fix Corda-Interop-App not able to build in case of github packages.


core/network/corda-interop-app/build.gradle
```diff
@@ -67,6 +67,19 @@ allprojects {
                     password github_password
                 }
             }
+        } else if (file("$projectDir/../github.properties").exists()) {
```
VRamakrishna, line 70: Why do we need this?

.github/workflows/deploy_corda-pkgs.yml
```diff
@@ -0,0 +1,106 @@
+name: Deploy Protos Java-Kotlin
```
VRamakrishna, line 1: Why not publish other packages too? Why just for Corda?

.github/workflows/deploy_docs.yml
```diff
@@ -31,6 +31,8 @@ jobs:
         working-directory: docs
         
       - name: Deploy
-        run: npm run deploy-gh-pages
-        working-directory: docs
+        uses: JamesIves/github-pages-deploy-action@v4.3.0
```
VRamakrishna, line 34: Hopefully this remains stable?

.github/workflows/test_asset-exchange-fabric.yml
```diff
@@ -89,7 +89,7 @@ jobs:
           fi
         working-directory: samples/fabric/fabric-cli
 
-  asset-exchange-fabric-local:
+  asset-exchange-fabric-semi-local:
```
VRamakrishna, line 92: Why do we need this? Which part is non-local here?

core/network/corda-interop-app/build.gradle
sanvenDev: Done.

.github/workflows/deploy_docs.yml
sanvenDev: That repo had many stars, so I guess it will be.

.github/workflows/test_asset-exchange-fabric.yml
sanvenDev: Semi-local, where we use local interopcc, but with go.mod instead of go.mod.local.
That's the only difference, thus it will catch any issue with interopcc when it will be published.

VRamakrishna (approved):

PR #259, opened 2022-05-12
Support for end-to-end confidentiality in the data sharing protocol

VRamakrishna:
Updated RFC specification:
- https://github.com/VRamakrishna/weaver-dlt-interoperability/blob/main/rfcs/models/security/confidentiality.md
- https://github.com/VRamakrishna/weaver-dlt-interoperability/blob/main/rfcs/models/security/confidentiality-design-choices.md

Implemented the feature for Fabric networks according to the above specification: common protos, Fabric Interoperation Chaincode, Fabric Node SDK, Relay.

Tested the feature by running integrated local tests: Fabric CLI, Fabric testnets.

Updated the tutorial documentation (and also reformatted notes in various pages):
- To view the rendered docs, see https://vramakrishna.github.io/weaver-dlt-interoperability/docs/external/introduction


rfcs/models/security/confidentiality.md
```diff
@@ -25,20 +25,24 @@ As the figure below illustrates, the data sharing protocol first involves a roun
```
dhinakaran2705, line 25: "Confidentiality through Asymmetric Cryptography" -> "Confidentiality and Authenticity through Asymmetric Cryptography"

dhinakaran2705, line 25: "The client application, which is the receiver of the data, can also be privy to the view contents as it cannot tamper with the data without invalidating the accompanying proof." -> Do you want to say how the client app will use the data here rather than saying that the client app can see the data just because it can't tamper with it?

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+
+- RFC: 01-008-appendix
+- Authors: Venkatraman Ramakrishna, Dhinakaran Vinayagamurthy, Sandeep Nishad, Krishnasuri Narayanam, Sikhar Patranabis
+- Status: Proposed
+- Since: 10-May-2022
+
+
+## Summary
+
+To fulfil the requirements described in the [end-to-end confidentiality spec](./confidentiality.md), the view contents returned by the source ledger must contain some unencrypted portions in the signed payload that can validate the plaintext contents as decrypted by the destination client. This allows the destination peers to validate the plaintext against tampering by the client. The solution for this was to add a hash of some form within the interoperation payload in the view.
```
dhinakaran2705, line 16: "the view contents returned by the source ledger must contain some unencrypted portions in the signed payload that can validate the plaintext contents as decrypted by the destination client." -> 
the view contents returned by the source ledger must contain some authentication of the payload that can validate the plaintext contents as decrypted by the destination client.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+
+## General Considerations
+
+For private (confidential) portions of a view's contents:
+- They should be encrypted with the receiver's public key.
+- If they only need tamper protection, they should be hashed.
+- If they need tamper protection and need to be authenticated by the receiver, they should be hashed and signed.
+For public (non-confidential) portions of a view's contents:
```
dhinakaran2705, line 24: Add in a new line.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+- If they need to be authenticated by the receiver, they should be signed.
+- If they only need tamper protection, they can be either hashed or signed.
+(_Note_: "hashed" implies "HMAC")
```
dhinakaran2705, line 27: You might want to say that we use HMAC for provable security guarantees. Or you could skip the HMAC reference altogether here.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+```
+M' = Decrypt(C)
+```
+DIM computes:
+```
+Verify: Sigma == Sign(P)
+(C', H) = Parse(P)
+Verify: H == SHA256Hash(M')
+```
+_Comments_: SHA256 is insecure as a hashing mechanism when confidentiality of M is required.
```
dhinakaran2705, line 65: No no.. SHA256 is a widely accepted "hash" mechanism. But a hash algorithm is not expected to provide confidentiality requirements. (Hash algorithms just provide one-wayness and collision resistance).

You could say that "This protocol provides authenticity of message to DIM, but the confidentiality of the message is not provided for the relays and some potential network eavesdroppers since hash doesn't provide ``semantic security'' guarantees that's expected out of an encryption scheme. Semantic security assures that confidentiality of the messages is provided even if the message space is finite and known to an adversary. Hash functions don't (and are not expected to) provide this property."

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+```
+Sigma = Sign(P)
+```
+DC computes:
+```
+M' = Decrypt(C)
+```
+DIM computes:
+```
+Verify: Sigma == Sign(P)
```
dhinakaran2705, line 61: This is not correct. Signatures are not (usually) deterministic.

You could say: "Signature verification algorithm on input Sigma and P outputs 1." Because we have abstracted out the verification key. If we want to have a simple equation type sentence here, we have to define Verify(vk, Sigma, P) (and Sign(sk, P)) before to use them here.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+```
+SP computes:
+```
+Sigma = Sign(P)
+```
+DC computes:
+```
+M' = Decrypt(C)
+```
+DIM computes:
```
dhinakaran2705, line 59: You might also want to say what all DIM gets from DC: M', P, Sigma.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+C = Encrypt(M || S)
+Sigma = Sign(C)
+```
+DC computes:
+```
+C' = Decrypt(C)
+```
+DIM computes:
+```
+Verify: Sigma == Sign(C)
```
dhinakaran2705, line 83: Same as in Protocol 1.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+DC computes:
+```
+C2 || H1 = Parse(P)
+C2` = Decrypt(C2)
+C1' || H' = Parse(C2')
+M' = Decrypt(C1')
+```
+DIM computes:
+```
+Verify: Sigma == Sign(P)
```
dhinakaran2705, line 115: Same comments as in Protocol 1.

rfcs/models/security/confidentiality-design-choices.md
```diff
@@ -0,0 +1,174 @@
...
+C || H = Parse(P)
+M1 = Decrypt(C)
+```
+DIM computes:
+```
+Verify: Sigma == Sign(P)
+M' || r' = Parse(M1)
+Verify: H == SHA256Hash(M' || r')
+```
+_Comments_: Security is dependent on the randomness of the sampled value, and the protocol requires no peer modification or custom signing logic.
```
dhinakaran2705, line 147: Give me a day. I will confirm the precise purpose of using HMAC instead of H. (Sometimes cryptographers suggest such usage to facilitate provable security guarantees without another version without having explicit attacks. But sometimes there could also exist some subtle attacks. Let me confirm what's the case here.)

rfcs/models/security/confidentiality.md
VRamakrishna: Done

VRamakrishna: No, there are unlimited ways in which a client app can use the data. In this context, all we are concerned about is that the client app cannot tamper with data in a way that beats its local IM's validation process. I've added a sentence to clarify.

rfcs/models/security/confidentiality-design-choices.md
VRamakrishna: Done.

rfcs/models/security/confidentiality-design-choices.md
VRamakrishna: Removed. I don't know why I added it in the first place.

rfcs/models/security/confidentiality-design-choices.md
VRamakrishna: Thanks. Added your text with some edits.

samples/fabric/fabric-cli/src/helpers/helpers.ts
```diff
@@ -832,19 +832,23 @@ const interopHelper = async (
       keyCert,
       false,
       options['relay-tls'] === 'true',
-      relayTlsCAFiles
+      relayTlsCAFiles,
+      options['e2e-confidentiality'] === 'true' ? true : false
```
sanvenDev, line 836: do we need tertiary operator here? Isn't the return value itself is boolean?

rfcs/models/security/confidentiality.md
```diff
@@ -25,20 +25,24 @@ As the figure below illustrates, the data sharing protocol first involves a roun
 
 <img src="../../resources/images/e2e-confidentiality.png" width=100%>
```
sanvenDev, line 26: image
This is how the image looks in dark mode, is it good? or do you want to edit it? (I use dark mode always, and thus noticing some texts slightly faint, its okay for me though). Also the backgrounds of hash, peer are not transparent, but white.

rfcs/models/security/confidentiality-design-choices.md
VRamakrishna: I was writing this informally. We don't mention keys or certificates anywhere and I didn't want to blow up the set of symbols. I think we can assume that keys and certificates exist for digital signatures without explicitly mentioning them.

VRamakrishna: Anyway, I'll change this simply to `Verify(Sigma, P)`.

rfcs/models/security/confidentiality-design-choices.md
VRamakrishna: I don't think this is necessary. What the DIM gets is indicated by what it can process (in the subsequent lines). I want to keep this minimal.

rfcs/models/security/confidentiality.md
VRamakrishna: This applies to all figures throughout this repository. We should submit a separate PR, adding a white background to every image.

VRamakrishna: Opened issue #260 to track.

samples/fabric/fabric-cli/src/helpers/helpers.ts
VRamakrishna: Done

sanvenDev (approved):

rfcs/models/security/confidentiality-design-choices.md
dhinakaran2705: Okay, this is good.

dhinakaran2705: Wait, a minor modification reqd. You would want to say, Verify(Sigma, P) == 1.

rfcs/models/security/confidentiality-design-choices.md
dhinakaran2705: Okay, it seems like we are mostly good. Just add a note that "Using HMAC enables provable security [1] of the authenticity property we strive for in our system."
with the citation: 
New Proofs for NMAC and HMAC: Security without Collision-Resistance, Mihir Bellare, Crypto 2006 eprint: https://eprint.iacr.org/2006/043.pdf

VRamakrishna: Done

PR #262, opened 2022-05-23
changes to support various hash mechanisms

knarayan:
Signed-off-by: Krishnasuri Narayanam <knaraya3@in.ibm.com>


common/protos/common/asset_locks.proto
```diff
@@ -19,18 +19,25 @@ message AssetClaim {
   bytes claimInfo = 2;
 }
 
+enum HashMechanism {
+  SHA256 = 0;
+  SHA512 = 1;
```
VRamakrishna, line 24: If we are including this as an option here, we should offer SHA512-based locking and claiming in the Interop Modules too.

core/network/fabric-interop-cc/libs/assetexchange/assetSwapContracts.go
```diff
@@ -417,14 +418,20 @@ func IsAssetLocked(ctx contractapi.TransactionContextInterface, callerChaincodeI
...
 	funName := "checkIfCorrectPreimage"
 	preimage, err := base64.StdEncoding.DecodeString(preimageBase64)
 	if err != nil {
 		return false, logThenErrorf("base64 decode preimage error: %s", err)
 	}
 
-	shaHashBase64 := GenerateSHA256HashInBase64Form(string(preimage))
+	shaHashBase64 := ""
+	if hashMechanism == common.HashMechanism_SHA256 {
+		shaHashBase64 = GenerateSHA256HashInBase64Form(string(preimage))
```
VRamakrishna, line 430: Add conditional for SHA512 here too.

core/network/fabric-interop-cc/libs/assetexchange/assetSwapContracts.go
```diff
@@ -1003,6 +1014,6 @@ func getHashPreImageHelper(ctx contractapi.TransactionContextInterface, key stri
 	if hashPreImageBase64Bytes == nil {
 		return "", logThenErrorf("key %s is not associated with any claimed asset", key)
 	}
-	
+
 	return string(hashPreImageBase64Bytes), nil
```
VRamakrishna, line 1018: Shouldn't we return a structure with the hash mechanism here, just like we do in `getHTLCHashHelper`?

common/protos/common/asset_locks.proto
knarayan: Done @VRamakrishna .

core/network/fabric-interop-cc/libs/assetexchange/assetSwapContracts.go
knarayan: Done @VRamakrishna .

core/network/fabric-interop-cc/libs/assetexchange/assetSwapContracts.go
knarayan: As discussed, since this is the preimage pertaining to just the hash mechanism ( and not generic to all lock mechanisms), we will not include the hash mechanism in the return value (and assume that the caller already knows this).

sanvenDev (approved):

PR #276, opened 2022-06-02
Go Modules Local Build Change

sanvenDev:
1. Change the way go modules are built locally (use `go mod vendor` instead of separate go.mod.local file).
2. Enable github packages workflow tests.
3. Add workflow to test build status of docker images of weaver.
4. Removed semi-local tests from workflows as local tests will be sufficient now.
5. Bump protos-java-kt to v1.3.1


.github/workflows/test_docker-build.yml
```diff
@@ -0,0 +1,116 @@
+name: Test All Docker Images Build
```
VRamakrishna, line 1: If this is a test for all Docker images in Weaver, we should include the relay images too, no?

.github/workflows/test_go.yml
```diff
@@ -17,91 +17,19 @@ jobs:
...
-
     - name: Build
       run: go build -v ./...
       working-directory: core/network/fabric-interop-cc/contracts/interop
 
     - name: Test
       run: go test -v ./...
       working-directory: core/network/fabric-interop-cc/contracts/interop
-
+
```
VRamakrishna, line 32: Remove whitespace

tests/network-setups/fabric/dev/scripts/setupAppCCLocal.sh
```diff
@@ -12,30 +12,8 @@ if [ -d "${CHAINCODE_PATH}/${CC_NAME}" ]; then
     echo "Deleting existing ${CC_NAME} chaincode folder to copy the latest..."
     rm -rf ${CHAINCODE_PATH}/${CC_NAME}
 fi
+
+(cd ${APP_CC_PATH} && make run-vendor)
```
VRamakrishna, line 16: Why do you need the parantheses?

tests/network-setups/fabric/dev/scripts/setupCCLocal.sh
```diff
@@ -5,34 +5,11 @@ INTEROP_CC_PATH=$PWD/../../../../core/network/fabric-interop-cc
...
-rm go.sum
-go mod tidy
-cd ../..
-mv go.mod.local go.mod
-rm go.sum
-go mod tidy
-cd ../..
-cp -R contracts/interop ../interop
-
+(cd $INTEROP_CC_PATH/contracts/interop && make run-vendor)
```
VRamakrishna, line 12: Why do you need the parantheses?

tests/network-setups/fabric/dev/scripts/setupAppCCLocal.sh
sanvenDev: parantheses makes sure when the command ends, we are back to original directory, and not to the one where did `cd` to.

tests/network-setups/fabric/dev/scripts/setupCCLocal.sh
sanvenDev: parantheses makes sure when the command ends, we are back to original directory, and not to the one where did cd to.

Issue #270, opened 2022-06-01
Manage event subscriptions and publications in Fabric Driver in source mode


VRamakrishna:

knarayan:
Here is the interface for event lookup:
```
async function lookupEventSubscriptions(eventMatcher: eventsPb.EventMatcher): Promise<Array<Query>> {
}

sanvenDev:
Closing issue as is subscription and publication is handled on driver in source mode.

PR #294, opened 2022-07-11
Weaver Fabric Driver changes for events

knarayan:
Changes address (pertaining to Fabric Driver):

- Event subscription
- Event unsubscription
- Event listing


sanvenDev:
Subscription addressed from these issues #270 @271

sanvenDev (approved):

core/drivers/fabric-driver/makefile
```diff
@@ -18,28 +18,28 @@ build-local:
 		cp -r ../../../common/protos-js/peer ./protos-js/ && \
 		cp -r ../../../common/protos-js/relay ./protos-js/ && \
 		cp -r ../../../common/protos-js/package.json ./protos-js/
```
VRamakrishna, line 20: Should we add `identity` here too just for completion? May be needed at a later date.

core/drivers/fabric-driver/server/events.ts
```diff
@@ -0,0 +1,323 @@
...
+        return retVal;
+    } catch(error: any) {
+        console.error(`Error during delete: ${JSON.stringify(error)}`);
+        await db?.close();
+        throw new Error(error);
+    }
+}
+
+async function lookupEventSubscriptions(
+    eventMatcher: eventsPb.EventMatcher
```
VRamakrishna, line 239: Add a function that does exactly what this function does but takes as parameters: <channel-id, contract-id, func-name>.

You can extract most of the code from here into that function and just call that from here by extracting the above values from the `eventMatcher` parameter.

core/drivers/fabric-driver/makefile
knarayan: Done!

core/drivers/fabric-driver/server/events.ts
knarayan: Done (updated the above function itself so that matching based on EventMatcher properties can happen)!

PR #295, opened 2022-07-18
Transaction Listener Support in Fabric Driver

### LINKED ISSUE #269: Transaction/block event listener support in Fabric driver

VRamakrishna:
Added logic to the Fabric driver to start block or chaincode event listeners upon an event subscription request.
Minor upgrades, bug fixes, CLI augmentation, and documentation enhancements.
Fixes #269 and #270 .


samples/fabric/fabric-cli/src/commands/event/unsubscribe.ts
```diff
@@ -109,29 +108,32 @@ const command: GluegunCommand = {
...
-            { address: data.view_address, Sign: true },
-            keyCert
-        )
-        
-        if (response.getStatus() == EventSubscriptionState.STATUS.UNSUBSCRIBED) {
-            console.log("Event Unsubscription Success for requestId:", response.getRequestId())
-        } else {
-            console.log("Unknown error")
+    const data = JSON.parse(fs.readFileSync(filepath).toString())
+    for (let i = 0 ; i < data.length ; i++) {
```
sanvenDev, line 112: If we want to have unsubscribe for an array of subscriptions, then we need to take input an array of request_ids for all those subscriptions.

core/drivers/fabric-driver/server/events.ts
```diff
@@ -8,26 +8,43 @@ import ack_pb from '@hyperledger-labs/weaver-protos-js/common/ack_pb';
...
+        console.log(`Sending to the relay the eventSubscription error Ack: ${JSON.stringify(ack_send_error.toObject())}`);
+        // Sending the fabric state to the relay.
+        client.sendDriverSubscriptionStatus(ack_send_error, relayCallback);
+    } else {
         const ack_send = new ack_pb.Ack();
         if (newRequestId == requestId) {
             // event being subscribed for the first time
+
+            // the event listener logic follows here
+            await registerListenerForEventSubscription(call_request, network_name);
```
sanvenDev, line 43: need to put `handlePromise` on this function, and if there is an error in event subscription, need to send an error ack with an appropriate msg.
Success ack on Line 46, 47 should only be executed if register listener was successful.

core/drivers/fabric-driver/server/events.ts
```diff
@@ -55,15 +83,7 @@ function subscribeEventHelper(
...
-    });
-}
-
-function unsubscribeEventHelper(
-    call_request: eventsPb.EventSubscription,
-    client: events_grpc_pb.EventSubscribeClient
-) {
-    const newRequestId = call_request.getQuery()!.getRequestId();
-    deleteEventSubscription(call_request.getEventMatcher()!, newRequestId).then((deletedSubscription: eventsPb.EventSubscription) => {
+    } else {
```
sanvenDev, line 86: Don't we need to unregister the listener before sending successful unsubscription?

core/drivers/fabric-driver/server/listener.ts
```diff
@@ -0,0 +1,198 @@
...
+                // Get transaction chaincode ID
+                const chaincodeId = transaction.payload.chaincode_proposal_payload.input.chaincode_spec.chaincode_id.name;
+                if (transaction.payload.chaincode_proposal_payload.input.chaincode_spec.input.args.length > 0) {
+                    // Get transaction function name: first argument according to convention
+                    const chaincodeFunc = transaction.payload.chaincode_proposal_payload.input.chaincode_spec.input.args[0].toString();
+                    console.log('Trying to find match for channel', channelId, 'chaincode', chaincodeId, 'function', chaincodeFunc);
+                    // Find all matching event subscriptions stored in the database
+                    let eventMatcher = new events_pb.EventMatcher();
+                    eventMatcher.setEventType(events_pb.EventType.LEDGER_STATE);
+                    eventMatcher.setEventClassId('');       // Only match subscriptions where class ID is not specified
```
sanvenDev, line 51: Need `*` here right? instead of empty string.

core/drivers/fabric-driver/server/listener.ts
```diff
@@ -0,0 +1,198 @@
...
+                            let errorString: string = error.toString();
+                            if (!errorString.includes('LEVEL_LOCKED')) {    // Check if contention was preventing DB access
+                                throw(error);
+                            }
+                            await new Promise(f => setTimeout(f, 2000));    // Sleep 2 seconds
+                        }
+                    }
+                    // Iterate through the view requests in the matching event subscriptions
+                    eventSubscriptionQueries.forEach(async (eventSubscriptionQuery: query_pb.Query) => {
+                        console.log('Generating view and collecting proof for channel', channelId, 'chaincode', chaincodeId);
```
sanvenDev, line 70: Do you want to add `transaction func` in the log as well?

core/drivers/fabric-driver/server/server.ts
```diff
@@ -61,16 +60,26 @@ function mockCommunication(query: query_pb.Query) {
     }, 3000);
 }
 
-// Handles communication with fabric network and sends resulting data to the relay
-const fabricCommunication = async (query: query_pb.Query, networkName: string) => {
-    // Query object from requestor
-    console.log('query', query.getRequestId());
+// Package view and send to relay
+export function packageView(
```
sanvenDev, line 64: If we need to export these functions, then let put these functions in `utils.ts`? Same for below export functions.

core/drivers/fabric-driver/server/listener.ts
```diff
@@ -0,0 +1,198 @@
...
+                        // Trigger proof collection
+                        const [result, invokeError] = await handlePromise(
+                            invoke(
+                                eventSubscriptionQuery,
+                                networkName,
+                            ),
+                        );
+                        if (!invokeError) {
```
sanvenDev, line 78: I'm just wondering if there is an error should be sending that error message also to the subscriber?
Because Event occurred but due to error in query, we are not sending anything, what is something is wrong in query view address itself, and that needs to be rectified?

core/drivers/fabric-driver/server/listener.ts
VRamakrishna: Done

core/drivers/fabric-driver/server/server.ts
VRamakrishna: Moved the client functions to `utils.ts` and the view packaging function to `fabric-code.ts`.

core/drivers/fabric-driver/server/events.ts
VRamakrishna: Done

core/drivers/fabric-driver/server/listener.ts
VRamakrishna: In the case you mention, yes, it'd be useful to send some feedback to the subscriber. But chances are most of these errors are transient ones, local to the source network, and we may end up spamming the subscriber. And also, if there is something wrong with the view address, the matching logic in the event listener is not going to catch it, and will instead silently ignore all block/contract events.

Error management is something we can design for later? For now, let event publish be best-effort, and let us assume that it is the responsibility of the subscriber to do its due diligence before sending a subscription request.

core/drivers/fabric-driver/server/events.ts
```diff
@@ -36,15 +36,16 @@ async function subscribeEventHelper(
...
-            await registerListenerForEventSubscription(call_request, network_name);
-            // set the following ack_send message below, only if the event listener logic is successful; else set to an error message
-
-            ack_send.setMessage('Event subscription is successful!');
-            ack_send.setStatus(ack_pb.Ack.STATUS.OK);
+        if (newRequestId == requestId) {    // event being subscribed for the first time
+            // Start an appropriate type of event listener for this event subscription if one is not already active
+            const [listenerHandle, error] = await handlePromise(registerListenerForEventSubscription(call_request, network_name));
+            if (error) {
+                ack_send.setMessage('Event subscription succeeded but listener registration failed');
```
sanvenDev, line 43: Here can we also delete the subscription from database? and just say event subscription failed?
Otherwise what do user needs to do in case of this error? they can't subscribe again as db entry is there, so is it something that the driver would automatically try to register listener from db later?

samples/fabric/fabric-cli/src/commands/event/unsubscribe.ts
VRamakrishna: Done.

core/drivers/fabric-driver/server/events.ts
VRamakrishna: Done. I'm logging warnings instead of failing, because listeners being active even when subscriptions have been deleted from the database is not a critical error; it's just a noop.

core/drivers/fabric-driver/server/events.ts
sanvenDev: `Event subscription error: listener registration failed`
Can you update that message at line 43?

sanvenDev (approved):

PR #296, opened 2022-07-20
Added Event Publication to Relay

sanvenDev:
1. Implemented 2 endpoints in `EventPublish`, to handle event data from driver and remote-relay.
2. Added an endpoint to fetch all received states: `GetEventStates` and protobuf message for output.
3. Added sdk function for above endpoint, and sample command for it in fabric-cli.
4. Added sample command in fabric-cli to start express server to listen to events.
5. Added more status to `RequestState`, also added status `DELETED` for cleanup of db states in relay (not implemented yet, TODO).
6. (EDIT) Added unit tests in relay for event publish.


common/protos/common/events.proto
```diff
@@ -68,4 +69,16 @@ message EventPublication {
...
+}
+
+message EventStates {
+    repeated EventState states = 1;
+}
+
+// the payload that is used for the communication between the requesting relay
+// and its network
+message EventState {
+  common.state.RequestState state = 1;
```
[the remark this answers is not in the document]
sanvenDev: Whenever we use, It will be like `EventState.state` or `event_state.state`, if you strongly feel to change this name, I can change it (just will have to change it from multiple places within relay code).

core/relay/src/services/event_publish_service.rs
```diff
@@ -28,11 +34,334 @@ impl EventPublish for EventPublishService {
...
+                reply
+            }
+        }
     }
 
+}
+
+
+/// send_driver_state is run on the remote relay. Runs when the driver sends the
+/// state back to the remote relay or if there was an error making the
```
knarayan, line 116: Please update the comment

core/relay/src/services/event_publish_service.rs
```diff
@@ -28,11 +34,334 @@ impl EventPublish for EventPublishService {
...
+        state,
+        event_sub_state.event_publication_spec.expect("No Publication Spec stored in event subscription state"),
+        request_id,
+        event_id.to_string(),
+        conf
+    );
+    
+    return Ok(());
+}
+
```
knarayan, line 278: Can you please add comment that it will either call the APP URL or call the Driver based on the subscriber of the event.

core/relay/src/services/event_subscribe_service.rs
```diff
@@ -1,14 +1,14 @@
...
-use crate::pb::common::events::EventSubscription;
+use crate::pb::common::events::{EventSubOperation, EventSubscription};
 use crate::pb::relay::events::event_subscribe_client::EventSubscribeClient;
 use crate::pb::relay::events::event_subscribe_server::EventSubscribe;
 // Internal modules
 use crate::db::Database;
 use crate::error::Error;
-use crate::relay_proto::{parse_address, LocationSegment};
-use crate::services::helpers::{update_event_subscription_status, get_driver, get_driver_client};
+use crate::relay_proto::{parse_address, LocationSegment}        ;
```
knarayan, line 10: Remote space.

core/relay/src/services/network_service.rs
```diff
@@ -48,6 +48,7 @@ impl Network for NetworkService {
                          );
                 match request_state.state.as_ref() {
                     Some(state) => {
+                        try_mark_request_state_deleted(request_state.clone(), request_id.to_string(), db);
```
knarayan, line 52: Add comment here that we delete only if the view was received and client also has got the view.

common/protos/common/events.proto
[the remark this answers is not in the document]
knarayan: Looks fine to me, no need of change I guess.

knarayan (approved):

PR #310, opened 2022-08-25
Basic code and documentation for Cross-Network Identity Sync Protocol

VRamakrishna:
Updated identity agent protobufs.
Added basic Fabric code to create wallet identities and perform smart contract operations.
Updated identity RFCs.


common/protos/identity/agent.proto
```diff
@@ -14,17 +15,38 @@ service IINAgent {
...
-  // user or agent triggers a flow to collect signatures attesting an
-  // external/foreign network unit's state and recording to ledger
-  rpc FlowAndRecordAttestations(NetworkUnitIdentity) returns (common.ack.Ack) {}
+  rpc SendIdentityConfiguration(AttestedMembership) returns (common.ack.Ack){}
   // Requesting attestation from a local IIN agent.
-  rpc RequestAttestation(NetworkUnitIdentity) returns (common.ack.Ack){}
+  rpc RequestAttestation(AttestedSecurityDomain) returns (common.ack.Ack){}
   // Handling attestation sent by a local IIN agent.
-  rpc SendAttestation(NetworkUnitIdentity) returns (common.ack.Ack){}
+  rpc SendAttestation(AttestedSecurityDomain) returns (common.ack.Ack){}
```
sanvenDev, line 22: We planned to send `Attestation` only right in this flow, not whole `AttestedSecurityDomain`?

common/protos/identity/agent.proto
```diff
@@ -14,17 +15,38 @@ service IINAgent {
...
 }
 
+// Unique identifier for a unit of a network that runs an IIN agent
 message NetworkUnitIdentity {
   string network_id = 1;
```
sanvenDev, line 27: Do you want to include a timestamp here (or somewhere else) or some id? so during a flow, that `id` can be used to track which particular request it is.
During async processing, the iin-agent will be stateless (i.e. not remember for which particular request it is getting response), and thus the information that is exchanged need to have some state/id information.
I thought of this structure because, this message is what is sent during the first step of the whole flow. So during first step only we can assign some timestamp or id.

core/identity-management/iin-agent/src/fabric-ledger/connector.ts
```diff
@@ -0,0 +1,59 @@
...
+            securityDomain: this.networkId,
+            members: memberships,
+        }
+        return securityDomainInfo;
+    }
+
+    // Invoke a contract to drive a transaction
+    // TODO: Add parameters corresponding to the output of a flow among IIN agents
+    async invokeContract() {
+        await invokeFabricChaincode(this.walletPath, this.connectionProfilePath, this.configFilePath, this.ledgerId, this.contractId, "", []);
```
sanvenDev, line 51: Do we not need to return anything here?

core/identity-management/iin-agent/src/fabric-ledger/connector.ts
```diff
@@ -0,0 +1,59 @@
...
+    configFilePath: string;
+    networkId: string;
+    orgMspId: string;
+    walletPath: string;
+
+    constructor(ledgerId: string, contractId: string, connectionProfilePath: string, configFilePath: string, walletPath: string) {
+        super(ledgerId, contractId);
+        this.connectionProfilePath = connectionProfilePath ? connectionProfilePath : path.resolve(__dirname, './', 'connection_profile.json');
+        this.configFilePath = configFilePath ? configFilePath : path.resolve(__dirname, './', 'config.json');
+        this.networkId = process.env.NETWORK_NAME ? process.env.NETWORK_NAME : 'network1';
```
sanvenDev, line 25: Should you take this networkId also as (optional) parameter, as all other variables are  taken as param instead of env variable?

core/identity-management/iin-agent/src/fabric-ledger/networkUtils.ts
```diff
@@ -0,0 +1,199 @@
...
+    configFilePath: string,
+    channelId: string,
+    chaincodeId: string,
+    functionName: string,
+    args: Array<string>,
+): Promise<any> {
+    console.log('Running invocation on Fabric channel and chaincode');
+    try {
+        const gatewayAndContract = await getNetworkContract(walletPath, connectionProfilePath, configFilePath, channelId, chaincodeId);
+        await gatewayAndContract.contract.submitTransaction(functionName, ...args);
```
sanvenDev, line 169: Same here, I think we should return whatever we get from submitTransaction? In case the invoked functions returns something.

core/network/fabric-interop-cc/contracts/interop/certificate_utils.go
```diff
@@ -112,7 +112,7 @@ func verifyCertificateChain(cert *x509.Certificate, certPEMs []string) error {
 				errMsg := fmt.Sprintf("Certificate link for Subject %s with Parent Subject %s invalid", caCert.Subject.String(), parentCert.Subject.String())
 				return errors.New(errMsg)
 			}
-			if i == len(certPEMs)-1 {
+			if i == len(certPEMs)-1 && cert != nil {
```
sanvenDev, line 115: this check may allow verification to be successful when cert is nil, i.e. that cert is not provided at all...

Also if you want to add it may be just check it when the function starts, so that we can throw error if the `cert` is nil, as we don't want to allow (success verification) if cert is nil. And here it will be assumed that cert is not nil.

core/network/fabric-interop-cc/contracts/interop/membership.go
```diff
@@ -96,8 +117,12 @@ func (s *SmartContract) GetMembershipBySecurityDomain(ctx contractapi.Transactio
 
 // verifyMemberInSecurityDomain function verifies the identity of the requester according to
 // the Membership for the external network the request originated from.
-func verifyMemberInSecurityDomain(s *SmartContract, ctx contractapi.TransactionContextInterface, cert *x509.Certificate, securityDomain string, requestingOrg string) error {
-	err := isCertificateWithinExpiry(cert)
+func verifyMemberInSecurityDomain(s *SmartContract, ctx contractapi.TransactionContextInterface, certPEM string, securityDomain string, requestingOrg string) error {
+	cert, err := parseCert(certPEM)
```
sanvenDev, line 121: Why do you want to pass string and parse cert here, given that all places that this function is being used they have already parsed the cert for some other uses. Would be redundant to parse it here also.

common/protos/identity/agent.proto
VRamakrishna: I'm not sure what we agreed on exactly, but upon reflection, I think it won't suffice to just send an `Attestation`. A secondary agent can't simply send an `Attestation` because it is attesting a collection (`AttestedMembershipSet`). So I think we can keep it as is and assume that the `AttestedSecurityDomain` parameter in this service endpoint will only contain the signature of the client.

common/protos/identity/agent.proto
VRamakrishna: Yes, I can add a `nonce` field. Or do you prefer `timestamp`?

core/identity-management/iin-agent/src/fabric-ledger/connector.ts
VRamakrishna: Yeah, I think we should return nil/error or equivalent (like what a Fabric chaincode function does).

core/network/fabric-interop-cc/contracts/interop/membership.go
VRamakrishna: The only reason for this change is to run the comparison in line 143. I agree with you about the redundancy otherwise. Let me know if there's a better way of doing things.

sanvenDev:
Update the version of fabric sdk in package.json (package-local.json) to 1.4.3.

ghoshbishakh (approved):
identity-syncing.md among others use old terms like Network DID instead of Security Domain DID. This can be fixed in later PRs.

sanvenDev (approved):

PR #314, opened 2022-09-09
Feature to select Endorsing Orgs in Fabric Data Sharing

sanvenDev:
1. Added list of orgs as parameter in `interopFlow` in SDK to chose endorsing orgs.
2. Added `participants` field under `ContractTransaction` in events.proto.
3. Updated fabric-driver `writeExternalState` endpoint to allow use of list of orgs to submit transaction to in case of event publication.
4. Updated the way to update docker images version using `VERSION` file (making it uniform with go modules).


common/protos/common/events.proto
```diff
@@ -62,6 +62,7 @@ message ContractTransaction {
   string func = 4;
   repeated bytes args = 5;
   uint64 replace_arg_index = 6;
+  repeated string participants = 7;
```
VRamakrishna, line 65: Can we call this `members` so we are consistent with terminology in the security domain and identity RFCs?

sanvenDev: okay

core/relay/src/client.rs
```diff
@@ -135,6 +135,7 @@ async fn event_suscribe(driver: bool) -> Result<(), Box<dyn std::error::Error>>
             func: "abc".to_string(),
             args: vec![],
             replace_arg_index: 0,
+            participants: vec![],
```
VRamakrishna, line 138: Change to `members`?

core/relay/src/client.rs
```diff
@@ -215,6 +216,7 @@ async fn event_unsuscribe(request_id: String, driver: bool) -> Result<(), Box<dy
...
```
VRamakrishna, line 219: Change to `members`?

VRamakrishna (approved):

PR #321, opened 2022-10-04
Go-sdk changes when confidential flag is set to false

knarayan:
Go-sdk changes when confidential flag is set to `false`.


core/relay/config/Fabric_Relay2.toml
```diff
@@ -32,6 +32,11 @@ hostname="localhost"
 port="9080"
 tls=false
 tlsca_cert_path="credentials/fabric_ca_cert.pem"
+[relays.Fabric_Relay2]
```
sanvenDev, line 35: Can you undo these changes, it may confuse people trying to read these samples config files as reference. Even I'd prefer this self reference to be removed from Fabric_Relay.toml also.

sanvenDev (approved):

sanvenDev:
Can you work on to write a workflow tests using go-cli and go-sdk as well? (in future)

VRamakrishna:
@knarayan I echo @sanvenDev 's suggestion to write workflow tests. We'd have detected the requirement for the changes you just made much earlier then. Make that the focus of your next PR, even before further augmentations in the Go SDK and CLI. Accompanied with that should be instructions in the tutorial docs. I believe it won't involve much work, as you'll just need to copy the existing fabric-cli-based workflows and make suitable switches to using go-cli.

VRamakrishna (approved):

VRamakrishna:
One more thing: re-enable the tests that are currently disabled before merging and make sure they pass: `data-sharing`, `fabric-asset-transfer`, and `asset-exchange-fabric`.

PR #322, opened 2022-10-07
IIN Agent Implementation of endpoints

sanvenDev:
IIN Agent:
1. Added implementation for endpoints `syncExternalState`, `requestIdentityConfiguration`, `sendIdentityConfiguration`, and `sendAttestation`.
2. Added utility functions in common/utils.
3. Added Config files (along with templates): dnsconfig, and security-domain-config.
4. Added env variables as required.
5. Added dockerfile and docker-compose for iin agent.
6. Added sample testnet envs for  iin agent.

SDK:
1. Added MembershipManager, to record local memberships and also perform manual sync of membership through iin-agent.

Fabric-CLI:
1. Removed generation of membership from all places.
2. Removed IINAgent user registration from cli.
3. Using SDK to record local membership.
4. Added function to sync membership manually via iin-agent (using SDK).

Driver:
1. Error Log fixes
2. Implemented function to restart event listeners from database when driver is restarted.
3. Allow TLSCACert for client to be empty.
4. Improved and slimmed docker image for fabric driver.

Workflows:
1. Updated workflows to use iin-agent for fabric membership sync.
2. Modified existing (different) workflow(s) to build, test and publish iin-agent docker image.
3. (EDIT) Added testcase for locally built docker based test net setup in data sharing workflow.

Update:
Updated all node applications to node version 16.17.1.


.github/workflows/test_asset-transfer.yml
```diff
@@ -556,6 +610,13 @@ jobs:
           ./bin/fabric-cli configure network --local-network=network2
           ./scripts/initAssetsForTransfer.sh
         working-directory: samples/fabric/fabric-cli
+                
+      - name: Fabric Sync Membership using IIN Agent
+        run: |
+          ./bin/fabric-cli configure membership --local-network=network1 --target-network=network2 --iin-agent-endpoint=localhost:9500
+          ./bin/fabric-cli configure membership --local-network=network2 --target-network=network1 --iin-agent-endpoint=localhost:9501
+          sleep 10
```
VRamakrishna, line 618: Is 10 enough to sync?

core/drivers/fabric-driver/server/events.ts
```diff
@@ -323,6 +323,32 @@ async function lookupEventSubscriptions(
     }
 }
 
+async function readAllEventMatchers(): Promise<Array<eventsPb.EventMatcher>> {
+    var returnMatchers = []
+    let db: DBConnector;
+    console.debug(`start readAllEventMatchers()`);
+    try {
+        // Create connection to a database
+        db = new LevelDBConnector(DB_NAME!);
```
VRamakrishna, line 332: Why don't you pass this as a function parameter, so that this function can be reused to read different databases on demand?

core/drivers/fabric-driver/server/listener.ts
```diff
@@ -291,9 +281,26 @@ const unregisterListenerForEventSubscription = async (
...
-    // TODO
+const loadEventSubscriptionsFromStorage = async (networkName: string): Promise<boolean> => {
+    console.log('Starting event listeners for subscribed events in database...')
+    try {
+        const eventMatchers = await readAllEventMatchers();
+        for (const eventMatcher of eventMatchers) {
+            try {
+                const listenerHandle = await registerListenerForEventSubscription(
+                    eventMatcher,
+                    networkName
```
VRamakrishna, line 292: Referring to my earlier comment: I suggest you pass the database type and name as parameter here.

.github/workflows/test_asset-transfer.yml
sanvenDev: It did pretty fast in my VM (within a sec or two), may be because there's only one node on each side. So I thought 10 should be sufficient.

core/drivers/fabric-driver/server/events.ts
sanvenDev: Currently I did the way it was done in other functions.
I think the idea was database can be configured (type and path) when driver is started based on env variables. After that all functions would refer to same database.

VRamakrishna: I get your point. Database is assumed to be Level DB everywhere, right? I guess that's something we can abstract away later then.

core/identity-management/iin-agent/README.md
```diff
@@ -33,3 +33,43 @@ make build
 ## Running the IIN Agent
 
 Run `npm run dev` to start the agent in your terminal. (_Note_: we will support a containerized IIN agent in the future.)
```
VRamakrishna, line 35: Shouldn't you have this section after the configuration section?

core/identity-management/iin-agent/README.md
```diff
@@ -33,3 +33,43 @@ make build
...
+{
+    "securityDomainName": "<ledgerId>",
+    "securityDomainName": [
+        "<node1/org1>",
+        "<node2/org2>"
+    ]
+}
+```
+
+It is JSON, with specifying who are the members of a security domain, with keys being the security domain.
```
VRamakrishna, line 52: The above structure is a bit misleading. You should put `securityDomainName` in angle brackets (`<>`) to indicate that it needs to be replaced by a name. Also, you should have just one key-value entry in the above example, with the possibilities for the value being listed below rather than in the structure.

core/identity-management/iin-agent/README.md
```diff
@@ -33,3 +33,43 @@ make build
...
+* For Fabric, value will be the channel name.
+* For Corda, value will be an array of nodes' endpoints. (Not implemented yet)
+
+### DNS Config
+
+Discovery is not implemented yet, so DNS details are added during bootstrapping. A sample `dnsconfig.json` looks like this:
+```
+{
+    "securityDomainName": {
```
VRamakrishna, line 61: Same point as above.

core/identity-management/iin-agent/dnsconfig.json.template
```diff
@@ -0,0 +1,14 @@
+{
+    "securityDomainName": {
```
VRamakrishna, line 2: Same point as above.

core/identity-management/iin-agent/README.md
```diff
@@ -33,3 +33,43 @@ make build
...
+        "iin-agent1-name": {
```
VRamakrishna, line 62: I think this is also not a keyword but something to be replaced with a real name, right?

core/identity-management/iin-agent/docker-compose.yml
```diff
@@ -0,0 +1,44 @@
...
+      - DNS_CONFIG_PATH=/opt/iinagent/dnsconfig.json
+      - SECURITY_DOMAIN_CONFIG_PATH=/opt/iinagent/security-domain-config.json
+      - SYNC_PERIOD=${SYNC_PERIOD}
+      - AUTO_SYNC=${AUTO_SYNC}
+    volumes:
+      - ${DLT_SPECIFIC_DIR}:/opt/iinagent/extra
+      - ${CONFIG_PATH}:/opt/iinagent/config.json
+      - ${DNS_CONFIG_PATH}:/opt/iinagent/dnsconfig.json
+      - ${SECURITY_DOMAIN_CONFIG_PATH}:/opt/iinagent/security-domain-config.json
+      - ../../relay/credentials:/opt/iinagent/credentials
```
VRamakrishna, line 31: I know this is a copy from the Fabric driver's compose file, but it was a hack there to begin with. Can we replace the prefix with an environment variable just like in the preceding lines? That way the creation of this container won't be tied to a hardcoded path for the relay's configuration.

And I think we should make the same change in the Fabric driver (and other places where this is hardcoded.)

core/identity-management/iin-agent/security-domain-config.json.template
```diff
@@ -0,0 +1,7 @@
+{
+    "securityDomainName": "<ledgerId>",
```
VRamakrishna, line 2: Same suggestions as earlier.

core/identity-management/iin-agent/README.md
sanvenDev: It could be an array of strings (for Corda, its array of nodes) or just a string (for Fabric, its channel name), so I'm not sure how to explain both in one key-value entry?

core/identity-management/iin-agent/README.md
```diff
@@ -33,3 +33,43 @@ make build
...
+        },
+        "iin-agent2-name": {
+            "endpoint": "hostname:port",
+            "tls": true,
+            "tlsCACertPath": "<cacert-path>"
+        }
+    }
+}
+```
+For each security domain, there's a JSON object which contains elements with key as iin-agent's name, which can be Org MSP Id for Fabric, or Node Id for Corda, and value as another JSON object. This value contains `endpoint` for the iin-agent, boolean `tls` which is true if tls is enabled for that iin-agent, and `tlsCACertPath` which specified the path to file containing tls CA certs, keep it empty string if not known.
```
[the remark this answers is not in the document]
sanvenDev: ", keep it empty string if not known", I actually meant this. I realized during Samara deployment of drivers, they didn't know what TLS cert to use on client side, so keeping it empty worked even when TLS was enabled (not sure how).

core/identity-management/iin-agent/src/common/utils.ts
```diff
@@ -14,3 +30,238 @@ export function handlePromise<T>(promise: Promise<T>): Promise<[T?, Error?]> {
...
+
+export function getLedgerBase(securityDomain: string, memberId: string): LedgerBase {
+    const ledgerId = getLedgerId(securityDomain);
+    if(!process.env.DLT_TYPE) {
+        throw new Error(`Env DLT_TYPE not defined`);
+    }
+    const dltType = process.env.DLT_TYPE!.toLowerCase();
+    if(dltType == 'fabric') {
+        const ledgerBase = new FabricConnector(ledgerId, process.env.WEAVER_CONTRACT_ID, securityDomain, process.env.CONFIG_PATH);
+        if (ledgerBase.orgMspId != memberId) {
```
VRamakrishna, line 76: Why do we pass `memberId` as parameter if its only purpose is to pass this check? Or, this parameter ought to be passed to the constructor too, so that the returned value stored in `ledgerBase` is a connection to the Fabric ledger for that particular org.

core/identity-management/iin-agent/src/common/utils.ts
```diff
@@ -14,3 +30,238 @@ export function handlePromise<T>(promise: Promise<T>): Promise<[T?, Error?]> {
...
+ * message: string
+ * certificate: pem string
+ * signature: base64 string
+ * returns: True/False
+ **/
+export function verifySignature(message, certificate, signature) {
+    const messageBuffer = Buffer.from(message);
+    const signBuffer = Buffer.from(signature, 'base64');
+    const publicKey = crypto.createPublicKey(certificate).export({type:'spki', format:'pem'});
+    return crypto.verify("SHA256", messageBuffer, publicKey, signBuffer);
```
VRamakrishna, line 149: Same comment as in the sign function.

core/identity-management/iin-agent/src/common/utils.ts
```diff
@@ -14,3 +30,238 @@ export function handlePromise<T>(promise: Promise<T>): Promise<[T?, Error?]> {
...
+ * Verifies a signature over message using SHA256
+ * message: string
+ * certificate: pem string
+ * signature: base64 string
+ * returns: True/False
+ **/
+export function verifySignature(message, certificate, signature) {
+    const messageBuffer = Buffer.from(message);
+    const signBuffer = Buffer.from(signature, 'base64');
+    const publicKey = crypto.createPublicKey(certificate).export({type:'spki', format:'pem'});
```
VRamakrishna, line 148: Does the "type" matter here?

core/identity-management/iin-agent/src/common/utils.ts
```diff
@@ -14,3 +30,238 @@ export function handlePromise<T>(promise: Promise<T>): Promise<[T?, Error?]> {
...
+    const certificate = attestation.getCertificate();
+    const message = membershipSerialized64 + nonce;
+    if (!verifySignature(message, certificate, signature)) {
+        console.error(`Error: Fail to verify signature on membership`);
+        return false;
+    }
+    return verifyMemberInMembership(membership, attestation);
+}
+
+export function verifyMemberInMembership(membership: membership_pb.Membership, attestation: agent_pb.Attestation) {
```
VRamakrishna, line 176: Indentations seem very inconsistent throughout this function.

core/identity-management/iin-agent/src/common/utils.ts
```diff
@@ -14,3 +30,238 @@ export function handlePromise<T>(promise: Promise<T>): Promise<[T?, Error?]> {
...
+            }
+            parentCert = caCert;
+        }
+		leafCACertPEM = chain[chain.length - 1];
+		isSignerRoot = (chain.length == 1);
+	}
+	return validateCertificateUsingCA(certificate, leafCACertPEM, isSignerRoot);
+}
+
+function validateCertificateUsingCA(cert: string, signerCACert: string, isSignerRootCA: boolean): boolean {
```
VRamakrishna, line 209: Check indentations as above.

core/identity-management/iin-agent/src/fabric-ledger/config.json.template
```diff
@@ -11,5 +11,8 @@
     },
     "mspId":"Org1MSP",
     "ordererMspIds": ["OrdererMSP"],
-    "caUrl":""
+    "ccpPath": "path-to-connection-profile",
```
VRamakrishna, line 14: Put angular brackets around the value. That's the general convention we use in other places whenever the given value is meant to be a variable rather than a constant.

core/identity-management/iin-agent/src/common/utils.ts
sanvenDev: We can not enforce connecting to particular org by `memberId`, it depends upon the connection profile.
The memberId is passed yes just to make sure that this is correct (intended) iin-agent.

E.g.: This is used when iin-agent attests membership requested by foreign network, so just to be sure it is actually the attester that requestor requested.

core/identity-management/iin-agent/src/fabric-ledger/connector.ts
```diff
@@ -5,44 +5,98 @@
...
+    iinAgentUserName: string;
 
-    constructor(ledgerId: string, contractId: string, networkId: string, connectionProfilePath: string, configFilePath: string, walletPath: string) {
-        super(ledgerId, contractId);
-        this.connectionProfilePath = connectionProfilePath ? connectionProfilePath : path.resolve(__dirname, './', 'connection_profile.json');
+    constructor(ledgerId: string, contractId: string, networkId: string, configFilePath: string) {
+        const weaverCCId = contractId ? contractId : 'interop';
+        super(ledgerId, weaverCCId);
+        //this.connectionProfilePath = connectionProfilePath ? connectionProfilePath : path.resolve(__dirname, './', 'connection_profile.json');
+        //this.walletPath = walletPath ? walletPath : path.join(process.cwd(), `wallet-${this.networkId}`);
```
VRamakrishna, line 31: Remove commented out lines.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
```diff
@@ -5,26 +5,224 @@
...
+    requestingSecurityDomain.setSecurityDomain(securityDomain);
+    requestingSecurityDomain.setMemberId(memberId);
+    
+    const request = new iin_agent_pb.SecurityDomainMemberIdentityRequest();
+    request.setRequestingNetwork(requestingSecurityDomain);
+    request.setNonce(nonce);
+    
+    // Request Attested Membership from all members of remote security_domain
+    const remoteSecurityDomainDNS = utils.getSecurityDomainDNS(remoteSecurityDomain);
+    let c = 0;
```
VRamakrishna, line 48: Call this `count` or something descriptive, as this variable is not just a loop variable.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
```diff
@@ -5,26 +5,224 @@
...
+    for (const iinAgent in remoteSecurityDomainDNS) {
+        const iinAgentClient = utils.getIINAgentClient(remoteSecurityDomain, iinAgent, remoteSecurityDomainDNS);
+        remoteSecurityDomainUnit.setMemberId(iinAgent);
+        request.setSourceNetwork(remoteSecurityDomainUnit);
+        console.log(`Requesting attested memberships from: ${remoteSecurityDomain} - ${iinAgent}`);
+        iinAgentClient.requestIdentityConfiguration(request, utils.defaultCallback);
```
VRamakrishna, line 54: Do we know at this point if the request succeeded, and can we catch errors here, like the server being down or the connection being broken? Or are you assuming that an error will be thrown? If it's the latter, then maybe you should catch it so you can store the right value in Line 58?

core/identity-management/iin-agent/src/protocols/externalOperations.ts
```diff
@@ -5,26 +5,224 @@
...
+        // remoteSecDomMapKeys.push(getSecurityDomainMapKey(remoteSecurityDomain, iinAgent, nonce));
+        c++;
+    }
+    foreignAgentResponseCount.set(nonce, { current:0, total: c});
 };
 
-// Generates security domain unit's state/configuration
+// Generates network unit's state/configuration
```
VRamakrishna, line 61: Wasn't the earlier label more accurate?

core/identity-management/iin-agent/src/common/utils.ts
sanvenDev: From docs it needs to be provided for public keys, quoting from docs
`Must be one of 'pkcs1' (RSA only) or 'spki'.`
I'm not sure what each of those mean, I tried `spki` and it was working.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
```diff
@@ -5,26 +5,224 @@
...
+        if (membership === "") {
+            membership = attestedMembershipOrError.getMembership();
+        } else if (membership !== attestedMembershipOrError.getMembership()) {
+            errorMsg = `received different membership from ${remoteAgent}`;
         }
+        if (!utils.validateAttestedMembership(membership, nonce, attestation)) {
+            errorMsg = `attested membership from ${remoteAgent} invalid.`;
+        }
+        attestations.push(attestation);
+        securityDomainMap.set(getSecurityDomainMapKey(remoteSecurityDomain, remoteAgent), attestedMembershipOrError);
```
VRamakrishna, line 157: Why do we need to manage these memory maps at the sending side (I know why we need them in the receiver)? From what I can see, they are never looked up; only set or deleted.

core/identity-management/iin-agent/src/protocols/localOperations.ts
```diff
@@ -14,6 +15,121 @@ export const requestAttestation = async (counterAttestedMembership: iin_agent_pb
 };
 
 // Processes attestations on a foreign security domain unit's state received from a local IIN agent
-export const sendAttestation = async (counterAttestedMembership: iin_agent_pb.CounterAttestedMembership) => {
-    console.log('sendAttestation:');
+export const sendAttestation = async (counterAttestedMembership: iin_agent_pb.CounterAttestedMembership, securityDomain: string, memberId: string) => {
+    // Assuming only one counter attestation in the list
```
VRamakrishna, line 19: Why?

core/identity-management/iin-agent/src/server.ts
```diff
@@ -147,13 +195,19 @@ if (process.env.IIN_AGENT_TLS === 'true') {
         configSetup().then(() => {
             console.log('Starting IIN agent server with TLS on', process.env.IIN_AGENT_ENDPOINT);
             iinAgentServer.start();
+            loopSyncExternalState();
```
VRamakrishna, line 203: Aren't we supposed to call this only if some environment variable is set?

core/identity-management/iin-agent/src/server.ts
```diff
@@ -130,9 +155,32 @@ iinAgentServer.addService(iin_agent_pb_grpc.IINAgentService, {
...
+    const delayTime: number = parseInt(process.env.SYNC_PERIOD ? process.env.SYNC_PERIOD : '3600');
+    const securityDomain = process.env.SECURITY_DOMAIN ? process.env.SECURITY_DOMAIN : 'network1';
+    const memberId = process.env.MEMBER_ID ? process.env.MEMBER_ID : 'Org1MSP';
+    const flagSync = process.env.AUTO_SYNC === 'false' ? false : true
+    while (flagSync) {
+        const secDomDNS = getAllSecurityDomainDNS();
+        for (const securityDomain in secDomDNS) {
+            const foreignSecurityDomain = new iin_agent_pb.SecurityDomainMemberIdentity();
+            foreignSecurityDomain.setSecurityDomain(securityDomain);
+            syncExternalStateFromIINAgent(foreignSecurityDomain, securityDomain, memberId);
```
VRamakrishna, line 183: Given how our protocol works, this doesn't need to be in a loop. It can be called exactly once, on one of the selected local IIN Agents. That agent can then orchestrate collection of signatures from other agents, which in turn will sync from external agents after seeing that no memberships are stored in memory.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
sanvenDev: Oh I didn't change this, may have been due to git merge probably retained old change. Will make it security domain.

core/identity-management/iin-agent/src/protocols/localOperations.ts
sanvenDev: When some local iinAgent counter attests, it sends only its attestation, so that's why assuming only one.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
sanvenDev: This is for caching the attested membership, so that if some local agent requests counter attestation, this agent can look in cache if it had recently fetched.

That code Shikhar will be writing, its not in this PR.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
sanvenDev: Yes in callback we'd know if request succesfully reached foreign iin agent or not.
But even if there's error, the logic would be to not proceed right? Because we want all foreign iin agent's attestations if some agent is down then should we still proceed by reducing the required attestations count?

core/identity-management/iin-agent/test/utilsTest.js
```diff
@@ -0,0 +1,163 @@
...
+const { Wallets } = require("fabric-network");
+const { ContractImpl } = require("fabric-network/lib/contract");
+const { NetworkImpl } = require("fabric-network/lib/network");
+const membershipPb = require('@hyperledger-labs/weaver-protos-js/common/membership_pb');
+const agent_pb = require('@hyperledger-labs/weaver-protos-js/identity/agent_pb');
+
+const utils = require('../out/common/utils.js')
+
+function decodeMembershipJSON(membershipJSON) {
+
```
VRamakrishna, line 28: Why is this blank?

core/network/fabric-interop-cc/contracts/interop/membership.go
```diff
@@ -39,15 +38,12 @@ func validateMemberCertChains(membership *common.Membership) error {
 }
 
 // Validate 'identity.Attestation' object against a message byte array
+// Should be removed?: to be confirmed in PR
```
VRamakrishna, line 41: This function isn't getting used, but the same pattern is used in `func validateAttestationsList`. So, if you want, you can refactor the latter to call this function.

core/network/fabric-interop-cc/contracts/interop/membership.go
```diff
@@ -180,15 +213,15 @@ func (s *SmartContract) CreateLocalMembership(ctx contractapi.TransactionContext
...
+func (s *SmartContract) UpdateLocalMembership(ctx contractapi.TransactionContextInterface, membershipSerialized64 string) error {
 	// Check if the caller has network admin privileges
 	if isAdmin, err := wutils.IsClientNetworkAdmin(ctx); err != nil {
 		return fmt.Errorf("Admin client check error: %+v", err)
 	} else if !isAdmin {
 		return fmt.Errorf("Caller not a network admin; access denied")
 	}
-
-	membership, err := decodeMembership([]byte(membershipJSON))
+
```
VRamakrishna, line 223: Remove whitespace.

sdks/fabric/interoperation-node-sdk/src/InteroperableHelper.ts
```diff
@@ -271,6 +271,12 @@ const signMessage = (message, privateKey) => {
     sign.end();
     return sign.sign(privateKey);
 };
+/**
+ * Verifies a signature over message using SHA256
+ **/
+const verifySignature = (message, publicKey, signature) => {
+    return crypto.verify("SHA256", message, publicKey, signature);
```
VRamakrishna, line 278: Take `"SHA256"` as an `algorithm` function parameter.

sdks/fabric/interoperation-node-sdk/src/MembershipManager.ts
```diff
@@ -0,0 +1,210 @@
...
+    securityDomain: string,
+    channelName: string,
+    weaverCCId: string
+): Promise<any> {
+    const network = await gateway.getNetwork(channelName)
+    const contract = network.getContract(weaverCCId)
+    return await contract.submitTransaction("DeleteLocalMembership", securityDomain);
+}
+
+function getMembershipUnit(channel: Channel, mspId: string): membership_pb.Member {
```
VRamakrishna, line 74: Can we import this and the functions below into the IIN Agent rather than leave redundant code? We can do the same for the Fabric driver later.

core/identity-management/iin-agent/src/protocols/externalOperations.ts
```diff
@@ -5,26 +5,224 @@
  */
```
VRamakrishna, line 5: Can you update the `README` file in this folder too? Minimal info.

core/identity-management/iin-agent/README.md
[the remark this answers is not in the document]
VRamakrishna: "; leave this empty if client-side TLS is not required, i.e., clients' certificates are trusted by default."

core/identity-management/iin-agent/src/common/utils.ts
VRamakrishna: There's a redundancy here. We specify the MSP ID in both the `config.json` and in the `.env` and then try to cross-check them here, which serves no purpose. Can we specify it just in one place?

core/identity-management/iin-agent/src/protocols/externalOperations.ts
VRamakrishna: I think there's a case for being selective with our retries; i.e., if a request to one remote IIN Agent fails, we shouldn't abandon and redo all other such requests. But this will probably make the code very complex, so I think we can revisit this later. The logic you currently have seems to fulfil the desired function.

VRamakrishna:
Why are the `fabric-asset-transfer` and `data-sharing` workflows disabled?

core/identity-management/iin-agent/src/common/utils.ts
sanvenDev: Okay can do that, but I didn't made this way to cross-check `.env` with `config.json` instead,
I did this because in case when this iin-agent receives `requestIdentity` request from remote network, it just passes the received `memberId` to this function, and this function checks whether received `memberId` in parameter is same as in config.json. 

Right now checking `memberId` is ledger depenedent i.e. orgMspId for `FabricConnector` something else for CordaConnector probably. So that's why to verify that I had it here instead of the place where we call this function (didn't want to put ledger specific switch case there).

I think better way would be to have `memberId` as property of `ledgerBase`, then this above check can easily be done in the respective endpoint function instead of here in utils, what do you think?

core/identity-management/iin-agent/src/server.ts
sanvenDev: I didn't exactly get your point. This is in loop, to sync memberships from all foreign networks defined in `dnsconfig.json`.

core/identity-management/iin-agent/test/utilsTest.js
sanvenDev: At some point I thought to create such a function but then created another (`getForeignAttestedMembership`), then forgot to remove this.
Now removed.

core/network/fabric-interop-cc/contracts/interop/membership.go
sanvenDev: The reason I didn't used this function is because, `cert` is already parsed in `validateAttestationsList`, now this function takes string and parse the cert again which seemed redundant.
If we remove parsing from this function, then all that remains is decode base64 signature and call `validateSignature`.

So that's why wasn't sure which one is better. Do you prefer this (remove parsing of cert from this function and take parsed certificate as parameter instead) way, or current way is better?

sanvenDev:
> Why are the `fabric-asset-transfer` and `data-sharing` workflows disabled?

Because of the changes in `interopcc` these test cases will fail. That's why.
I changed the way local membership is recorded, now am passing serialized protobuf. Its easy to generate this and pass it from SDK, rather than JSON.

core/identity-management/iin-agent/src/common/utils.ts
VRamakrishna: Yeah, I agree. Member ID is an attribute in our protos, so it's meant to be DLT-neutral. Keeping it in `ledgerBase` should work.

core/identity-management/iin-agent/src/server.ts
VRamakrishna: Ah....sorry, I misread what you were looping on. I thought you were iterating over the orgs.

core/network/fabric-interop-cc/contracts/interop/membership.go
VRamakrishna: Why not remove `parseCert` from `validateAttestationsList` and retain it in this function instead? If you need the result of the parsing for some other purpose, you can add it to the return value of this function.

sanvenDev: Done.

core/identity-management/iin-agent/src/common/utils.ts
sanvenDev: Done.

VRamakrishna (approved):

PR #324, opened 2022-10-13
IIN Agent Auto Sync Bug and Deploy Workflow improvement

sanvenDev:
* IIN Agent auto sync bug fixed. Now works as expected.
* Added more config instructions in README of IIN Agent
* Deploy workflows now will pass if package to be published already exists.


core/identity-management/iin-agent/README.md
```diff
@@ -70,6 +70,65 @@ Discovery is not implemented yet, so DNS details are added during bootstrapping.
 ```
 For each security domain, there's a JSON object which contains elements with key as iin-agent's name, which can be Org MSP Id for Fabric, or Node Id for Corda, and value as another JSON object. This value contains `endpoint` for the iin-agent, boolean `tls` which is true if TLS is enabled for that iin-agent, and `tlsCACertPath` which specifies the path to file containing TLS CA certs, keep it empty string if not known.
 
+### Environment Variables
+
+Following are the list of environment variables for IIN Agent:
```
VRamakrishna, line 75: Add that these need to be specified in the `.env`, whose template file is `.env.template`.

VRamakrishna (approved):

core/identity-management/iin-agent/README.md
sanvenDev: Done.

PR #329, opened 2022-10-25
protobuf refactor: type to assettype

michiboo:
refactor type to AssetType in protobuf lock and its related used places


common/protos-solidity/common/asset_locks.proto
```diff
@@ -0,0 +1,69 @@
+syntax = "proto3";
```
sanvenDev, line 1: This file should not be checked in, but instead should be generated from the file in `protos folder`

common/get-solidity-protos.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+WEAVER_ROOT=${1:-".."}
+
+BESU_PROTOSDIR=${WEAVER_ROOT}/common/besu-protos
```
VRamakrishna, line 5: Whatever you are downloading or creating in this folder is not specific to Besu, right? So I'd suggest renaming this folder and moving it as well as this script file into `common/protos-solidity`. This logic doesn't need to be in the root `common` folder.

core/identity-management/iin-agent/README.md
```diff
@@ -34,7 +34,7 @@ make build
```
VRamakrishna, line 34: Are these `iin-agent` changes showing up here because of a merge? Because these look like changes @sanvenDev made earlier.

common/protos/common/asset_locks.proto
```diff
@@ -28,27 +28,28 @@ message AssetLockHTLC {
   HashMechanism hashMechanism = 1;
   bytes hashBase64 = 2;
   uint64 expiryTimeSecs = 3;
-  enum TimeSpec {
+  TimeSpec timeSpec = 4;
+}
+
+enum TimeSpec {
```
VRamakrishna, line 34: Given that we've moved `TimeSpec` to the `common` namespace, I wonder if we should name it `LockTimeSpec`, because it is specific to how a lock's time is represented.

common/get-solidity-protos.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
sanvenDev: and don't use `besu-protos`, re-use `protos-solidity`

core/network/besu/contracts/interop/.gitignore
```diff
@@ -0,0 +1 @@
+*.proto
```
sanvenDev, line 1: I realise that you won't be copying `.proto` files here but `.sol` files of compiled protos right? So may be you can copy the compiled protos into one directory say `protos-sol`, and add that directory to `.gitignore`?

sanvenDev (approved):

PR #340, opened 2022-12-04
create besu sdk

michiboo:
Signed-off-by: micky <chanmickyyun@gmail.com>


core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -57,18 +58,65 @@ contract AssetExchangeContract is ERC1155Holder {
...
+            uint256 c = uint256(uint8(b[i]));
+            if (c >= 48 && c <= 57) {
+                result = result * 10 + (c - 48);
+            }
+        }
+    }
+
     function lockAsset(
-        address receiver,
+        bytes memory rawParams,
```
sanvenDev, line 94: may be call this `assetExchangeAgreement`? Also do you want to remove amount parameter as well? Because if its fungible lock, this parameter would contain `FungibleAssetExchangeAgreement`.

For Hybrid cases, I thought we discussed to declare new protobuf message called `HybridAssetExchangeAgreement` where both AssetID and quantity fields would be present.

May be have a flag as input to indicate which case this is (i.e. whether fungible, non-fungible, or hybrid, to decide to which protobuf to decode)

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -87,18 +135,20 @@ contract AssetExchangeContract is ERC1155Holder {
...
-                sender,
-                receiver,
-                assetContract,
-                amount,
-                hashLock,
-                expirationTime
-            )
-        );
+        if (hashMechanism == HashMechanism.SHA256) {
+            lockContractId = sha256(
```
sanvenDev, line 141: to generate lockContractID, you don't need to depend upon HashMechanism. HashMechanism is just for Hash Algo used in HTLC hash generation.
This lockContractID is independent of that and is just for book-keeping. So you can always have sha256 for this and remove the if condition check here.

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -57,18 +58,65 @@ contract AssetExchangeContract is ERC1155Holder {
...
+        AssetLockHTLC memory assetInfo;
+        (, , assetInfo) = AssetLockHTLCCodec.decode(
+            0,
+            lockinfo,
+            uint64(lockinfo.length)
+        );
+        HashMechanism hashMechanism = assetInfo.hashMechanism;
+        bytes32 hashLock = bytesTobytes32(assetInfo.hashBase64);
+        if (hashMechanism == HashMechanism.SHA256) {
+            hashLock = sha256(abi.encodePacked(hashLock));
```
sanvenDev, line 115: I don't get what are you trying to do here, `hashLock` itself is the hash, why are you computing hash of the hash?

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -140,17 +190,29 @@ contract AssetExchangeContract is ERC1155Holder {
     }
 
     // The receiver claims the ownership of an asset locked for them once they obtain the preimage of the hashlock
-    function claimAsset(bytes32 lockContractId, bytes32 preimage)
+    function claimAsset(bytes32 lockContractId, bytes memory lockInfoProtobuf)
```
sanvenDev, line 193: call this param `claimInfoProtobuf` or `claimInfoSerializedBase64`?

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -140,17 +190,29 @@ contract AssetExchangeContract is ERC1155Holder {
...
+            lockInfoProtobuf,
+            uint64(lockInfoProtobuf.length)
+        );
+        bytes32 preimage = bytesTobytes32(
+            slice(params.hashPreimageBase64, 2, 32)
+        );
         // Check the validity of the claim
         require(c.status == LOCKED, "lockContract is not active");
         require(block.timestamp < c.expirationTime, "lockContract has expired");
+        bytes32 encryptedPreimage;
```
sanvenDev, line 210: Its not encrypted pre-image, may be call this as computedHash? as in we are now computing hash using the preimage and comparing it with provided hash in LockContract?

sdks/besu/interoperation-node-sdk/build/AssetManager.js
```diff
@@ -0,0 +1,205 @@
+"use strict";
```
sanvenDev, line 1: Don't commit the build directory, add it to `.gitignore` here. This can be generate by running `make build` right?

sdks/besu/interoperation-node-sdk/src/AssetManager.ts
```diff
@@ -0,0 +1,164 @@
...
+    senderAddress: string,
+    preimage: Buffer, // pass hash
+): Promise<any> => {
+
+    if (!assetManagerContract) {
+        console.log("Contract Address not supplied");
+        return false;
+    }
+    
+    const hash = new SHA256()
```
sanvenDev, line 107: when you take this `hash` as param, you don't need to then have this code (currently this means only SHA256 is used, but we want user to be in control of choosing hash mechanism)

sdks/besu/interoperation-node-sdk/src/AssetManager.ts
```diff
@@ -0,0 +1,164 @@
...
+    return {result: lockStatus};
+};
+
+/**
+ * Latter step of a Hashed Time Lock Contract
+ * - Claim a unique asset instance using a hash preimage
+ **/
+const claimAssetInHTLC = async (
+    lockContractId: string,
+    assetManagerContract: any,
```
sanvenDev, line 97: Just for consistency with other SDK functions in this file, can you move `assetManagerContract` parameter as the first parameter?

sdks/besu/interoperation-node-sdk/src/AssetManager.ts
```diff
@@ -0,0 +1,164 @@
...
+        console.log(`Recipient address not supplied ${recipientAddress}`);
+        return {result: false};
+    }
+    const currTimeSecs = Math.floor(Date.now() / 1000);   // Convert epoch milliseconds to seconds
+    if (expiryTimeSecs <= currTimeSecs) {
+        console.log("Supplied expiry time invalid or in the past: %s; current time: %s", new Date(expiryTimeSecs).toISOString(), new Date(currTimeSecs).toISOString());
+        return {result: false};
+    }    
+
+    const protobufParams = createAssetExchangeAgreementSerialized("", String(assetID), recipientAddress.slice(2), Web3.utf8ToHex(String(assetData)));
```
sanvenDev, line 71: Shouldn't the last parameter be `senderAddress`? as that is the locker.
what is `assetData` exactly?
Also the first parameter should not be empty, have something in the asset type, take it as another function parameter and pass it here.

Also based on comment on contract, this will change and depending upon the case you should either create this, or fungible asset exchange agreement or hybrid one.

core/network/besu/contracts/interop/asset_locks.proto.sol
```diff
@@ -906,6 +906,293 @@ library AssetExchangeAgreementCodec {
 
 }
 
+struct HybridAssetExchangeAgreement {
```
sanvenDev, line 909: Don't commit this file, can you delete this and add it in gitignore?
This is generated after building protos right?

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -87,8 +138,209 @@ contract AssetExchangeContract is ERC1155Holder {
...
+            receiver,
+            assetContract,
+            amount,
+            hashLock,
+            expirationTime,
+            lockContractId
+        );
+    }
+
+    function lockAsset(
```
sanvenDev, line 294: Call this lockHybridAsset

core/network/besu/contracts/interop/manageAssetAny.sol
```diff
@@ -140,17 +392,25 @@ contract AssetExchangeContract is ERC1155Holder {
...
+        );
+        bytes32 preimage = bytesTobytes32(
+            slice(params.hashPreimageBase64, 2, 32)
+        );
         // Check the validity of the claim
         require(c.status == LOCKED, "lockContract is not active");
         require(block.timestamp < c.expirationTime, "lockContract has expired");
         require(
-            c.hashLock == sha256(abi.encodePacked(preimage)),
+            c.hashLock == preimage,
```
sanvenDev, line 413: why this change? hashLock should not be equal to preimage, hashLock should be hash of preimage.

Also can you add if else conditions here to check the HashMechanism, and depending upon HashMechanism use sha256 only if HashMechanism is SHA256, else throw error?

sdks/besu/interoperation-node-sdk/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
sanvenDev, line 1: Don't commit `package-lock.json`, delete it and add it to gitignore.

sdks/besu/interoperation-node-sdk/src/AssetManager.ts
```diff
@@ -0,0 +1,222 @@
...
+        return { result: false };
+    }
+    const currTimeSecs = Math.floor(Date.now() / 1000);   // Convert epoch milliseconds to seconds
+    if (expiryTimeSecs <= currTimeSecs) {
+        console.log("Supplied expiry time invalid or in the past: %s; current time: %s", new Date(expiryTimeSecs).toISOString(), new Date(currTimeSecs).toISOString());
+        return { result: false };
+    }
+    var protobufParams;
+    if (assetType == "ERC721") {
+        protobufParams = createAssetExchangeAgreementSerialized("", String(assetID), recipientAddress.slice(2), Web3.utf8ToHex((assetData)));
```
sanvenDev, line 92: Instead of if-else like this can you add 3 functions? createFungibleHTLC, createNonFungibleHTLC, createHTLC, similar to `lockFungibleAsset`, `lockNonFungibleAsset`, and `lockAsset` in contract?

sanvenDev:
Closing this PR as #348 is opened new with the updates as per the comments

PR #352, opened 2022-12-21
Add instructions to enable asset exchange for Fabric

### LINKED ISSUE #131: Adapt Fabric network for asset exchange
Fill "Asset exchange" section in https://labs.hyperledger.org/weaver-dlt-interoperability/docs/external/getting-started/enabling-weaver-network/fabric

### LINKED ISSUE #172: View Generation in Fabric Driver should handle differing responses from peers
The Fabric driver, as a client of a Fabric network, seeks endorsements from several peers as per a verification policy. The [view structure](https://github.com/hyperledger-labs/weaver-dlt-interoperability/blob/main/rfcs/formats/views-fabric.md#view-data-definition) presently supports only one copy of the peers' responses, under the assumption that the responses will agree. A remote client/network can, of course, validate the signatures against the response to determine if the verification policy was satisfied, but there may be scenarios in which the remote client/network may wish to make use of differing endorsements from remote peers. We need to check if adding multiple responses is necessary or useful, and if so, update the view structure and the Fabric driver implementation.

sanvenDev:
- Closes #131 
- Updated Fabric_View protobuf to handle multiple proposal_response_payload in driver and chaincode.
- Bumped proto versions.
- Fixed corda interop app to work with updated Fabric_View protobuf.
- Bumped corda modules versions.
- Driver updated to handle multiple proposal_response_payloads from different peers, and package in Fabric_View. Closes #172 

Doc deployed here: https://sanvendev.github.io/weaver-dlt-interoperability/docs/external/getting-started/enabling-weaver-network/fabric/#development


core/drivers/fabric-driver/server/fabric-code.ts
```diff
@@ -161,28 +159,37 @@ async function invoke(
 
         // 4. Prepare the view and return.
         const viewPayload = new view_data.FabricView();
-        const endorsements: proposalResponse.Endorsement[] = [];
+        const endorsedProposalResponses: view_data.FabricView.EndorsedProposalResponse[] = [];
         //TODO Fix ts error
         //@ts-ignore
+        let ii = 0;
```
VRamakrishna, line 165: If we are going to leave this in the code, rename this to be more descriptive, like `endorsementCounter`.

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
```diff
@@ -34,103 +34,240 @@ Let us walk through the changes that are required in different phases of your ne
...
-        logger.info('Getting MSP Info for org: ' + org + '.');
-        let mspConfig = network.getChannel().getMsp(mspId);
-        let securityGroupMspConfig = {};
-        securityGroupMspConfig.chain = [];
-        if (Array.isArray(mspConfig.rootCerts)) {
-            for (let i = 0; i < mspConfig.rootCerts.length ; i++) {
-                securityGroupMspConfig.chain.push(mspConfig.rootCerts[i]);
+- _Chaincode_: 
+  - _Data Sharing_: No code changes are required for Weaver enablement, as mentioned above.
+  - _Asset Exchange_: Asset exchange can be enabled in two ways, either use weaver `assetexchange` as library, or use interop chaincode to handle asset exchange by doing a chaincode to chaincode call.
```
VRamakrishna, line 39: Wrap `assetexchange` with a link to the library package, here and in similar references below.

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
```diff
@@ -34,103 +34,240 @@ Let us walk through the changes that are required in different phases of your ne
...
+            ...
+            return true, nil
         }
-        res.setHeader('Content-Type', 'application/json')
-        res.send(security_group_config);
-    });
+        ```
+  - _Asset Transfer_: _TBD_
+
+- _Layer-2 applications_: Let us examine the adaptations required in detail:
```
VRamakrishna, line 235: Let's call this "Client (or Layer-2) Applications", just to avoid confusion.

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
```diff
@@ -353,6 +558,112 @@ To start the driver, navigate to the folder containing the above files and run t
 docker-compose up -d
 ```
 
+#### Launch IIN Agents
+
```
VRamakrishna, line 562: Do we need all of these instructions below just to start IIN Agents? This looks like a dump of the README, and quite intimidating. On this page, we should give as many shortcuts as possible (pull this image, configure env variables as follows, etc.), and then link to the README for details.

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
```diff
@@ -404,17 +715,114 @@ To prepare your network for interoperation with a foreign network, you need to r
...
-  As in the above two cases, use `CreateMembership` to record a confiuration for the first time for a given `securityDomain` and `UpdateMembership` to overwrite a configuration.
-
-  | Notes |
-  |:------|
-  | Security domain configurations (organization lists and their certificate chains) for any Fabric network channel are subject to change, so you should run the above procedure periodically in a loop. |
+- **Local network security domain (membership) configuration**:
+  Call the function or HTTP endpoint to record local membership that was created in development step.
 
 Your Fabric network is now up and running with the necessary Weaver components, and your network's channel's ledger is bootstrapped with the initial configuration necessary for cross-network interactions!
+
```
VRamakrishna, line 722: Below lines are all commented out, and I saw them earlier in the instructions. can you just delete them?

core/drivers/fabric-driver/server/fabric-code.ts
sanvenDev: Yeah I thought for debugging purpose would be good to have those logs for future

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
sanvenDev: Yeah wanted to ask, currently I've only listed all chaincode and API functions with `contractId`.
Others I skipped but had it in the comment, in case you think we should add those also. Its just it'd make the page too long, and probably they won't need to know all of them (only one of them either all contractIds or without contractIds)

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
sanvenDev: It is similar to what you are suggesting (its same as what's in driver section), it is long because there are 3 config files required and then set of environment variables.

sanvenDev: I think it would be good to have it here, rather than pointing them to README, so that docs can standalone be enough to explain whatever is required.

VRamakrishna: Can we provide instructions then in a less dry manner? Start with, to launch a single IIN agent, pull container image XYZ, edit the following configurations files in these ways, etc. Then say "To launch an additional IIN Agent corresponding to a different network member (organization in Fabric, node in Corda, etc.), repeat the above steps but change these specific things...". Something like this.

Try to edit this section a little bit along these lines. I can reorganize content in a later PR.

docs/docs/external/getting-started/enabling-weaver-network/fabric.md
VRamakrishna: OK. Then further above, just add a line saying that there is another API that an app developer can use, which doesn't involve contract IDs. For details, point them to the READMEs.

sanvenDev: Okay moved those commented text in README, and added a link here.

core/network/fabric-interop-cc/libs/assetexchange/README.md
```diff
@@ -3,4 +3,194 @@
...
+```
+require(
+    ...
+    github.com/hyperledger-labs/weaver-dlt-interoperability/common/protos-go v1.5.3
+    github.com/hyperledger-labs/weaver-dlt-interoperability/core/network/fabric-interop-cc/libs/assetexchange v1.5.3
+    ...
+)
+```
+
+There are two kinds of APIs of library, one with `contractId`, and other without `contractId`.
```
VRamakrishna, line 18: Replace with: `The asset exchange library within the Weaver chaincode package offers two sets of APIs to manage asset locks and claims for a cross-ledger asset exchange. In one set, the transactions take a unique contract ID referring to an asset exchange instance as parameter. This contract ID is generated by the Weaver chaincode when the first asset is locked (as per the protocol steps), and returned to the calling application client. Though you must use contract ID-based API functions for fungible assets, you may also use them for non-fungible assets. For non-fungible assets, Weaver also provides a parallel set of API functions that take asset type and ID as parameters rather than a temporarily generated contract ID.`

core/network/fabric-interop-cc/libs/assetexchange/README.md
```diff
@@ -3,4 +3,194 @@
...
+
+## With ContractId
```
VRamakrishna, line 20: Can we have consistent naming? Is `ContractId` or `contractId` a keyword? If yes, we should use that everywhere with markdown. Otherwise, let's just say "contract ID" in plain English.

sanvenDev: This was a heading so that's why used ContractId here, everywhere else we have used `contractId`

VRamakrishna (approved):

core/network/fabric-interop-cc/libs/assetexchange/README.md
sanvenDev: Its not a keyword btw, its just our defined variable name. But also want to avoid confusion of contract ID with smart contract ID

PR #372, opened 2023-02-01
Enhancement for Asset Manager SDK in Fabric

sanvenDev:
Added `endorsingOrgs` parameter in AssetManager SDK transaction functions. (kept query functions as it is)


sdks/fabric/interoperation-node-sdk/src/AssetManager.ts
```diff
@@ -126,12 +127,17 @@ const createHTLC = async (
     //If timeoutCallback is not defined, start automated listener
     if (!timeoutCallback) 
     {
-        StartHTLCAssetLockListener(contract, "", assetType, assetID, recipientECert, "", assetLockExpirationCallback);
+        StartHTLCAssetLockListener(contract, "", assetType, assetID, recipientECert, "", assetLockExpirationCallback, endorsingOrgs);
```
VRamakrishna, line 130: We need to guard all these code additions by `if (endorsingOrgs && Array.isArray(endorsingOrgs) && endorsingOrgs.length > 0)`.

sanvenDev: I've set default parameter as empty array `[]` and type of argument as `Array<string>`, still do you think we need the 1st two checks? 
and the place where it is being used there we have check `endorsingOrgs.length > 0`, here we are just passing it to the event listener and setting it, it can be empty or have elements, but when the event listener calls `reclaim`, the `reclaim` checks that array is not empty.

sanvenDev: I can change the condition at line 135 (and other similar places), to: 
`if (endorsingOrgs && endorsingOrgs.length > 0)`
But do you think we need guard over the start listener function?

VRamakrishna (approved):

PR #398, opened 2023-04-19
feat(fabric-driver): Added monitor for missed events

sandeepnRES:
1. Added Monitor in fabric driver for missed events.
2. Removed contract event listener, now block listener serves both block and contract events.
3. Two env variables to configure monitor:
```
MONITOR_ENABLE: Enable or disable monitor
MONITOR_SYNC_PERIOD: time period in seconds for monitor to sleep before running again
```
4. Fix Fabric testnet for Mac
5. Upgrade Fabric testnet to v2.5


core/drivers/fabric-driver/server/listener.ts
```diff
@@ -5,165 +5,179 @@
...
+                    TxId: ${txEvent.transactionId},
+                    TxStatus: ${txEvent.status},
+                    isTxValid: ${txEvent.isValid}\n`)
+                if (txEvent.isValid) {
+                    validTxIdSet.add(txEvent.transactionId);
+                    logger.debug(`BlockEvent Listener: Valid tx added: ${txEvent.transactionId}`);
                 }
-                await new Promise(f => setTimeout(f, 2000));    // Sleep 2 seconds
+            })
+            if (currBlockNum === lastBlockNum + 1) {
```
VRamakrishna, line 163: Can `currBlockNum - lastBlockNum` not be greater than 1? In which case, shouldn't we parse all of those blocks?

core/drivers/fabric-driver/server/server.ts
```diff
@@ -253,6 +253,22 @@ const configSetup = async () => {
...
+    const delayTime: number = parseInt(process.env.MONITOR_SYNC_PERIOD ? process.env.MONITOR_SYNC_PERIOD : '30');
+    const networkName = process.env.NETWORK_NAME ? process.env.NETWORK_NAME : 'network1';
+    const flagEnable = process.env.ENABLE_MONITOR === 'false' ? false : true;
+    if (flagEnable) {
+        logger.info("Starting monitor...");
+        logger.info("Monitor sync period: ", delayTime);
+    } else {
+        logger.info("Monitor disabled.");
+    }
+    while (flagEnable) {
```
VRamakrishna, line 266: Is there a way to disable this dynamically, i.e., via a service request?

tests/network-setups/.gitignore
```diff
@@ -4,6 +4,8 @@
 fabric/dev/config
 fabric/dev/bin
 fabric/dev/.fabric-setup
+fabric/dev/install-fabric.sh
```
VRamakrishna, line 7: Why have this?

tests/network-setups/fabric/dev/Makefile
```diff
@@ -100,7 +99,8 @@ deploy-cc-local: setup-cc-local
 
 
 .fabric-setup:
-	curl -sSL https://bit.ly/2ysbOFE | bash -s -- $(FABRIC_VERSION) $(FABRIC_CA_VERSION) $(FABRIC_COUCH_VERSION) -s
+	curl -sSLO https://raw.githubusercontent.com/hyperledger/fabric/main/scripts/install-fabric.sh && chmod +x install-fabric.sh
+	./install-fabric.sh --fabric-version $(FABRIC_VERSION) --ca-version $(FABRIC_CA_VERSION) docker binary
```
VRamakrishna, line 103: Looks like the `install-fabric.sh` script needs to be checked in, no?

core/drivers/fabric-driver/server/server.ts
sandeepnRES: Can be done by adding a new endpoint

tests/network-setups/.gitignore
sandeepnRES: the new way of downloading fabric binaries and images, is downloading this script and this script then downloads rest of stuff.
Hence added the script in gitignore, because this script will be downloaded everytime the testnet is started.

tests/network-setups/fabric/dev/Makefile
sandeepnRES: It is downloaded in line 102. These are the fabric steps to download images and binaries, don't want to put script created by Fabric people in codebase, its better to download everytime I think.

core/drivers/fabric-driver/server/listener.ts
sandeepnRES: This is listener code, and we decided that if difference is greater than 1, then listener will discard and let monitor parse all those blocks, so that there is no duplication of events.

sandeepnRES: Line 354 does that in Monitor function

VRamakrishna (approved):

PR #399, opened 2023-04-20
fix(driver): monitor to always process events

sandeepnRES:
Monitor should always process events if there is a single block not yet read.


VRamakrishna (approved):

VRamakrishna (changes requested):
I think you need to bump up the driver's version for a republish.

PR #404, opened 2023-05-22
Package Dependency Updates in Fabric Interop CC and Besu Samples

VRamakrishna:
- Updated 'solc' dependency in Besu 'simmpleasset' Truffle configuration to handle breaking change.
- Updated go-ethereum version in Fabric Interop CC to ensure that latest dependencies are used.
- Updated zlib version in Fabric Interop CC Dockerfile to avoid vulnerability.


sandeepnRES:
can you also update version in `contracts/interop/VERSION` to probably 1.5.14, so that new interopcc image gets published automatically after PR merge?

VRamakrishna:
> can you also update version in `contracts/interop/VERSION` to probably 1.5.14 (same as driver), so that new interopcc image gets published automatically after PR merge?

Done.

